PR #1, opened 2018-11-05
Feature/entities

Laurian:


pietrop:
Looking good, 

- I've dome some refactor to make some of the functions a bit more readable and easy to follow.
- moved the generate entities function to separate module so that can be reused by other adapters for other STT services.
- and added unit tests for generate entities function for troubleshooting and making sure it worked as expected

PR #8, opened 2018-11-09
Added support for transcripts without punctuation.

chrisbaume:


pietrop:
Since you mention this was for converting [Gentle](https://github.com/lowerquality/gentle) output to the BBC Kaldi format, and it was because it turns out Gentle strips punctuation,  I'd suggest making 2 adapters to support these as input for the Transcript Editor

1. Gentle Transcription JSON
2. Gentle Alignement JSON

For option 2, the issue is that some of the words don't have timings (eg if Gentle does not manage to align them) - however as you mentioned these could be interpolated, looking at the once before and after.

pietrop:
see https://github.com/bbc/react-transcript-editor/pull/10

pietrop:
Closing this issue, as BBC Kaldi will always expected to have punctuation, and we can add the adapters for Gentle as described above if needed.

PR #9, opened 2018-11-12
Media player controls

pietrop:


src/lib/Util/timecode-converter/index.js
```diff
@@ -0,0 +1,66 @@
...
+    if(typeof newCurrentTimeInSeconds !== 'number'){
+        return timecodes.fromSeconds(parseFloat(time))
+    }
+    return timecodes.fromSeconds(time)   
+}
+
+/**
+ * @param {*} time 
+ * Can take as input timecodes in the following formats 
+ * - hh:mm:ss:ms
```
bevand10: Hi Pietro.

In broadcast VIDEO, timecode is NOT hh:mm:ss:ms, it's hh:mm:ss:ff where ff is frames, dependent on the framerate of the media concerned.

So, users such as Panaroma that will be transcoding video media (or audio taken from video) will be expecting hh:mm:ss:ff.

See https://github.com/bbc/newslabs-cdn/blob/master/js/20-bbcnpf.utils.js for regularised BBC broadcast-compliant seconds2timecode functions.

pietrop: Awesome, thanks for flagging that, good catch. Having a look at your code, and seeing if I can extract the relevant functions to do those changes, will let you know when I have something to show

PR #18, opened 2018-11-22
Feature: Added custom block renderer, shows timecodes

jamesdools:


pietrop:
Made another branch to try out a few things 
https://github.com/bbc/react-transcript-editor/tree/render-timecodes-spike
from commit message

>can update the contenState, change a speaker name and then check the console.log you'd see it can change the entity map, but now the wrapperBlock needs to communicate the new state to TimedTextEditor so that it can set the state there as well and go full circle

PR #45, opened 2018-12-06
Export adapters

chrisbaume:
Method for allowing export of transcripts using adapters.


src/index.js
```diff
@@ -83,10 +84,20 @@ class App extends React.Component {
...
+    const {data, ext} = this.refs.transcriptEditor.exportData(this.state.exportFormat);
+    this.download(data, `${ this.state.mediaUrl } .${ext}`);
+  }
+
   getEditorContent = () => {
-    const tmpEditorsContnet = this.refs.transcriptEditor.getEditorContent(this.state.sttType);
+    const tmpEditorsContent = this.refs.transcriptEditor.getEditorContent(this.state.sttType);
 
-    this.download(JSON.stringify(tmpEditorsContnet, null, 2), `${ this.state.mediaUrl } .json`)
+    this.download(JSON.stringify(tmpEditorsContent, null, 2), `${ this.state.mediaUrl } .json`)
```
pietrop: I'd remove the space between the file name and the file extension

```suggestion
    this.download(JSON.stringify(tmpEditorsContent, null, 2), `${ this.state.mediaUrl }.json`)
```

[btw using the beta -  suggest a specific change to the line](https://help.github.com/articles/reviewing-proposed-changes-in-a-pull-request/)

src/lib/TranscriptEditor/TimedTextEditor/export-adapters/txt/index.js
```diff
@@ -0,0 +1,4 @@
+export default (blockData) => {
+  const lines = blockData.blocks.map(x => x.text);
+  return lines.join(' ');
```
pietrop: I'd add a line breaks to respect the paragraphs, if the user edits the text, and arranges it with the desired paragraph breaks, it be good to preserve this when exporting in plain text.

```suggestion
  return lines.join('\n\n');
```

When we add speaker names at paragraph level we can revisit this, and add more option to the plain text output eg optional speakers, and timecodes and paragraph level.

src/lib/TranscriptEditor/TimedTextEditor/export-adapters/index.js
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Adapters for Draft.js conversion
+ * @param {json} blockData - Draft.js blocks
+ * @param {string} exportFormat - the type of file supported by the available adapters
+ */
+
+const exportAdapter = (blockData, exportFormat) => {
+  switch (exportFormat) {
+    case 'draftjs':
+      return { data: JSON.stringify(blockData), ext: 'json' };
```
pietrop: This assumes that the users is going to export the json as plain text, and therefore is needed as a string, while they might want to do something else with the json programatically, eg add it into another draftJS instance for example or other operation... in any case happy to leave it as is for now and can be revisited later if the need arises.

meanwhile I would just add the pretty print option to `JSON.stringify` so that the output once downloadable is more readable.

```suggestion
      return { data: JSON.stringify(blockData, null,2), ext: 'json' };
```

pietrop:
Thanks @chrisbaume ! 
Looks good, just a few minor comments - see above.

I've used the [suggest a specific change to the line  beta feature](https://help.github.com/articles/reviewing-proposed-changes-in-a-pull-request/) so that they are easy to incorporate. 

Other then that will also just let @jamesdools have a look before we go ahead.

src/index.js
```diff
@@ -83,10 +84,20 @@ class App extends React.Component {
...
   }
 
+  handleExportFormatChange = (event) => {
+    console.log(event.target.name, event.target.value)
+    this.setState({ [event.target.name]: event.target.value });
+  }
+
+  exportTranscript = (exportFormat) => {
+    const {data, ext} = this.refs.transcriptEditor.exportData(this.state.exportFormat);
+    this.download(data, `${ this.state.mediaUrl } .${ext}`);
```
pietrop: same here, I'd remove the space between file name and file extension

```suggestion
    this.download(data, `${ this.state.mediaUrl }.${ext}`);
```

PR #53, opened 2018-12-17
Refactor

jamesdools:


src/lib/TranscriptEditor/TimedTextEditor/Word.js
```diff
@@ -2,29 +2,37 @@ import React, { PureComponent } from 'react';
...
 
 class Word extends PureComponent {
-  render() {
-    const data = this.props.entityKey
-      ? this.props.contentState.getEntity(this.props.entityKey).getData()
-      : {};
-
-    const confidence = data.confidence > 0.6 ? 'high' : 'low';
+  generateConfidence = (data) => {
+    return data.confidence > 0.6 ? 'high' : 'low';
```
pietrop: There's an edge case, when the confidence score it's not provided. we should think about how to handle that. coz with the current check, everything is then mark as low confidence, which is not very helpful. 

screen shot 2018-12-17 at 17 21 08

But yeah this could be a separate ticket.

src/lib/TranscriptEditor/MediaPlayer/PlayerControls.js
```diff
@@ -76,7 +76,11 @@ PlayerControls.propTypes = {
   timecodeOffset: PropTypes.string,
   promptSetCurrentTime: PropTypes.func,
   rollback: PropTypes.func,
-  handleMuteVolume: PropTypes.handleMuteVolume
+  handleMuteVolume: PropTypes.handleMuteVolume,
```
pietrop: Should probably be a function?

```suggestion
   handleMuteVolume: PropTypes.func,
```

Otherwise raises error 
`Warning: Failed prop type: PlayerControls: prop type `handleMuteVolume` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`.`

pietrop:
I'd suggest possibly merging PR https://github.com/bbc/react-transcript-editor/pull/54#issuecomment-447955137 into this branch, before merging to master, as it has a fix on the splitting paragraph issue. and fixing the merge conflict on that refactor shouldn't be too bad 🤞.

pietrop:
`defaultHotKeys.js` line 5 needs to be set to `handler: ( ) => { self.togglePlayMedia(); },` to enable keyboard shortcut

pietrop:
The latest commit of moving the play/pause in the state before changing the video element, might be causing performance issues when using `pauseWhileTyping`. 

When you start typing you would get a significant delay, where you won't see the text being updated on screen. until you stop typing. this is a new bug.

PR #54, opened 2018-12-17
Refactor handle key command - split-paragraph

pietrop:
some refactoring of the split-paragraph `handleKeyCommand` key bindings into smaller function to aid with legibility of the overall functionality.


pietrop:
also fixes issue from refactor PR https://github.com/bbc/react-transcript-editor/pull/53 where splitting a paragraph does not compute timecodes or speakers

screen shot 2018-12-17 at 19 46 49

PR #55, opened 2018-12-17
Refactor play media spike

pietrop:
Suggested refactor `playMedia` function to discuss a possible simpler implantation, where the function only takes one **optional** boolean argument.

Tested and it works on 
- [x] click on video preview
- [x] clicking on play/pause button
- [x] used react tool to test `isPausedWhileTyping` - true/false for `TimedTextEditor`
- [x] `esc` keyboard shortcut


pietrop:
renamed `togglePlayMedia` as `playMedia` , as couldn't find it used/declared anywhere else?

pietrop:
might not be needed.

PR #60, opened 2018-12-21
Add IBM Watson STT Json import support.

AndrewDAnderson:
No.

**Describe what the PR does**    
Adds initial support for IBM Watson TTS Json files. Working well in my limited tests, but a few things could probably be refactored for clarity. Creating EntityRanges is done from within the adapter itself, rather than being handed off to the common function for that purpose...

Ready to be reviewed.

**Additional context**    
N/A

### DIFF
--- a/src/lib/Util/adapters/ibm/index.js
+++ b/src/lib/Util/adapters/ibm/index.js
@@ -1,0 +1,266 @@
+/**
+ * Convert IBM json
+ ```
+ {
+   "created": "2018-12-18T22:40:38.903Z",
+   "id": "f14cbe1e-0315-11e9-b590-a3e1e3acdd73",
+   "updated": "2018-12-18T22:47:30.392Z",
+   "results": [{
+      "result_index": 0,
+      "speaker_labels": [
+         {
+            "speaker": 1,
+            "confidence": 0.36,
+            "final": false,
+            "from": 13.04,
+            "to": 13.23
+         }, ...]
+      "results": [
+         {
+            "final": true,
+            "alternatives": [{
+               "transcript": "there is a day about ten years ago when I asked a friend to hold a baby dinosaur robot upside down ",
+               "timestamps": [
+                  [
+                     "there",
+                     13.04,
+                     13.23
+                  ],
+                  [
+                     "is",
+                     13.23,
+                     13.36
+                  ] ...
+
+               "confidence": 0.938,
+               "word_confidence": [
+                  [
+                     "there",
+                     0.902
+                  ],
+                  [
+                     "is",
+                     0.52
+                  ],
+                  [
+                     "a",
+                     0.819
+                  ] ...
+        ...
+```
+ *
+ * into
+ *
+ ```
+ const blocks = [
+  {
+    "text": "There is a day.",
+    "type": "paragraph",
+    "data": {
+      "speaker": "TBC 0",
+      "words": [
+        {
+          "start": 13.02,
+          "confidence": 0.68,
+          "end": 13.17,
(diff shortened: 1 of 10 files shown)


src/lib/Util/adapters/ibm/index.js
```diff
@@ -0,0 +1,266 @@
...
+        const startTime = result.alternatives[0].timestamps[z][1];
+        const endTime = result.alternatives[0].timestamps[z][2];
+        const wordConfidence = result.alternatives[0].hasOwnProperty('word_confidence') ?
+          result.alternatives[0].word_confidence[z][1] : 100;
+
+        if (newSentence === true) {
+          punctuatedWord = capitalizeFirstLetter(word);
+          firstWordOfSentence = word;
+          newSentence = false;
+          data.speaker = 'TBC ' + topLvlResult.speaker_labels[z].speaker;
```
pietrop: IBM gives you speaker labels, and I've seen you are using that in the adapter, which is great
Only minor thing, I'd change the `TBC` to `Speaker`?

```suggestion
           data.speaker = 'Speaker ' + topLvlResult.speaker_labels[z].speaker;
```

As we have been (informally) using `TBC` as a place holder when the STT service doesn't return any speaker diarization info, and/or when it has not been incorporated in the adapter yet.

pietrop:
That's great, thanks @AndrewDAnderson 

I guess you meant to write STT (Speech To Text) and not TTS (Text To Speech) in the title of the PR (?).

Looks good, and thanks for adding tests!

A few comments 

1. As you mentioned Creating EntityRanges could be done with the shared helper function, unless there are any advantages in doing that within the IBM adapter?

2. Made a comment/suggestion re-speaker label names - see inline.

3. It might be good to have more meaningful names for `x`, `y`,`z`, I do realised they are just loop counters, for example `topLevelCounter`,`wordLevelCounter` and `sentenceLevelCounter` (?)

_We'll have another look and see if there's anything else, thanks_

src/lib/Util/adapters/ibm/index.js
AndrewDAnderson: Sounds great. I had originally had it as "Speaker" but didn't know if "TBC" had any... meaning and was trying to make this adapter 'fit in' as much as possible. ;-)

AndrewDAnderson:
> I guess you meant to write STT (Speech To Text) and not TTS (Text To Speech) in the title of the PR (?).

Yes. Note to self: do not make public commits at 3am.

> 1. As you mentioned Creating EntityRanges could be done with the shared helper function, unless there are any advantages in doing that within the IBM adapter?

Honestly, I didn't investigate the helper function at all - I had already started writing an adapter for a project and just tweaked it a little for use here. I do have questions about why there is so much duplication of data - but I will ask those in place.

> 2. Made a comment/suggestion re-speaker label names - see inline.

Responded inline. Good suggestion.

> 3. It might be good to have more meaningful names for `x`, `y`,`z`, I do realised they are just loop counters, for example `topLevelCounter`,`wordLevelCounter` and `sentenceLevelCounter` (?)

Sounds good. This was just a quick attempt to get it *functional* - and I'm sure others would like basic IBM support, so I was pushing it out the door with plans of cleaning things up later.

pietrop:
Hi @AndrewDAnderson , 
Just circling back to see if you are interested in re-basing this, and address some of the suggestions mentioned above so that we can merge it into master?

We just added support for Speechmatics https://github.com/bbc/react-transcript-editor/pull/94 and thought might be good to wrap up the IBM one as well.

Let me know

PR #66, opened 2019-01-03
help-box-spike: moved ToolTip next to settings icon

pietrop:
I think https://github.com/bbc/react-transcript-editor/pull/65 look good, but thought it might had been easier to show my suggestions then describe it etc..

screen shot 2019-01-03 at 19 42 09

- moved the `How Does this work` next to settings and keyboard shortcuts, this saves the extra line in `TimedTextEditor`.
- because after moving the tooltip the black background didn't quiet work with the background of the `MediaPlayer` / `PlayerControls` area, changed it to background colour of the keyboard shortcuts, better colour scheme match, readability, consistency etc..
- moved some of the CSS properties to the props of the component. 
- as refactor, to keep things tidy, I've noticed that the [`ToolTip` component has a list of CSS properties it takes as props](https://www.npmjs.com/package/react-simple-tooltip#props), might be good to make use of those as much as possible and add extras only when there might be shortcomings.

Let me know what you think

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3441,7 +3441,7 @@
         },
         "regjsparser": {
           "version": "0.1.5",
-          "resolved": "http://registry.npmjs.org/regjsparser/-/regjsparser-0.1.5.tgz",
+          "resolved": "https://registry.npmjs.org/regjsparser/-/regjsparser-0.1.5.tgz",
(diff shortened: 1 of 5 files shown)


pietrop:
Also see this note for some possible future work outside of the scope of this PR https://github.com/bbc/react-transcript-editor/projects/1#card-16169755

pietrop:
disregarded in favour of #65

Issue #71, opened 2019-01-04
Save to local storage does not work with local media (eg blobs).


pietrop:
Reason is that local storage save at the moment uses the url of file as the key, and for local media the blob has different url all the time.

needs some research into alternatives eg change key to file name and not url? (for all or for blob url only?)

pietrop:
Some notes of what I got so far. 

TL;DR: still no viable solution.

---

In the demo app once the local media is uploaded we create an object url and pass that to the component.

```js
const fileURL = URL.createObjectURL(file);
```

So I was looking to see if we could use the file name(as this would not change 🤞and should allow us to retrieve the corresponding transcript from local storage on page refresh)  instead of the blob url, without having to pass it to the component.

after a bit of exploration I had figured out that you can convert an object url back into a blob

```js
var blob = null;
var xhr = new XMLHttpRequest(); 
xhr.open("GET",  window.fileURL);  // `window.fileURL` is the url to convert
xhr.responseType = "blob";//force the HTTP response, response-type header to be blob
xhr.onload = function() 
{
    blob = xhr.response;//xhr.response is now a blob object
}
xhr.send();
```

However the problem is that a blob by its nature only has two attributes, `size` and `type`

```js
{size: 85047555, type: "video/mp4"}
```

as explained here on [stackoverflow](https://stackoverflow.com/questions/27159179/how-to-convert-blob-to-file-in-javascript) you could then convert this into a file
```js
var file = new File([blob], "name");
```

But it doesn't seem possible to get the original file name of an object url even after converting it to a blob as it is not stored in the blob.

pietrop:
fixed in this PR https://github.com/bbc/react-transcript-editor/pull/72

PR #72, opened 2019-01-04
fixed local media local storage issue

pietrop:

Related to Issue https://github.com/bbc/react-transcript-editor/issues/71   `Save to local storage does not work with local media (eg blobs).` 

**Describe what the PR does**    
 
added optional `fileName` param to use as key for saving  local media into local storage. 

Ready for review 

**Additional context**    

Only catch is that the mediaUrl should be present along side the transcriptData when initialising the component. Eg I've added some logic in demo app to make sure media is present when uploading transcript json.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,7 +18,8 @@ class App extends React.Component {
       mediaUrl: null,
       isTextEditable: true,
       sttType: 'bbckaldi',
-      analyticsEvents: []
+      analyticsEvents: [],
+      fileName: 'Kate Darling Ted Talk'
     };
     // this.handleChangeLoadTranscriptJson = this.handleChangeLoadTranscriptJson.bind(this);
   }
@@ -45,6 +46,7 @@ class App extends React.Component {
       this.setState({
         // transcriptData: kaldiTedTalkTranscript,
         mediaUrl: fileURL,
+        fileName: file.name
       });
     }
   }
@@ -58,22 +60,28 @@ class App extends React.Component {
     });
   }
 
-  handleChangeLoadTranscriptJson(files) {
-    const self = this;
-    const file = files[0];
-    // let type = file.type;
-    // TODO: add checks
-    // let transcriptJsonContent = FileReader.readAsText(file)
-    const fr = new FileReader();
-    fr.onload = function (e) {
-      // e.target.result should contain the text
-      console.log(JSON.parse(e.target.result));
-      self.setState({
-        transcriptData: JSON.parse(e.target.result),
-        // mediaUrl: tedTalkVideoUrl
-      });
-    };
-    fr.readAsText(file);
+  handleChangeLoadTranscriptJson(e) {
+    const files = e.target.files;
+    if (this.state.mediaUrl === null) {
+      e.preventDefault();
+      alert('fist add a video or audio file, and then re-import the transcription');
+    }
+    else {
+      const self = this;
+      const file = files[0];
+      // let type = file.type;
+      // TODO: add checks
+      // let transcriptJsonContent = FileReader.readAsText(file)
+      const fr = new FileReader();
+      fr.onload = function (e) {
+        // e.target.result should contain the text
+        console.log(JSON.parse(e.target.result));
+        self.setState({
+          transcriptData: JSON.parse(e.target.result)
+        });
+      };
+      fr.readAsText(file);
+    }
   }
 
   handleIsTextEditable = () => {
@@ -117,6 +125,9 @@ class App extends React.Component {
      this.setState({ analyticsEvents: [ ...this.state.analyticsEvents, event ] });
    }
 
+   handleChangeTranscriptName = (value) => {
+     this.setState({ fileName: value });
+   }
    render() {
      return (
        <div className={ style.container }>
@@ -133,18 +144,6 @@ class App extends React.Component {
          <br />
          <button onClick={ () => this.loadDemo() }>load demo</button>
          <hr />
-         <label>open Transcript Json</label>
-         <SttTypeSelect
-           name={ 'sttType' }
-           value={ this.state.sttType }
-           handleChange={ this.handleSttTypeChange }
-         />
-         <input
-           type="file"
-           onChange={ e => this.handleChangeLoadTranscriptJson(e.target.files) }
-         />
-
-         <br />
          <label>Load Local Media</label>
          <input
            type="file"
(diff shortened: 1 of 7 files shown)


src/index.js
```diff
@@ -58,22 +60,28 @@ class App extends React.Component {
...
+    const files = e.target.files;
+    if (this.state.mediaUrl === null) {
+      e.preventDefault();
+      alert('fist add a video or audio file, and then re-import the transcription');
+    }
+    else {
+      const self = this;
+      const file = files[0];
+      // let type = file.type;
+      // TODO: add checks
```
jamesdools: I think this might be an important thing to add in first 🤔

src/index.js
```diff
@@ -58,22 +60,28 @@ class App extends React.Component {
...
-        transcriptData: JSON.parse(e.target.result),
-        // mediaUrl: tedTalkVideoUrl
-      });
-    };
-    fr.readAsText(file);
+  handleChangeLoadTranscriptJson(e) {
+    const files = e.target.files;
+    if (this.state.mediaUrl === null) {
+      e.preventDefault();
+      alert('fist add a video or audio file, and then re-import the transcription');
```
jamesdools: typo :)

src/index.js
```diff
@@ -58,22 +60,28 @@ class App extends React.Component {
...
+    }
+    else {
+      const self = this;
+      const file = files[0];
+      // let type = file.type;
+      // TODO: add checks
+      // let transcriptJsonContent = FileReader.readAsText(file)
+      const fr = new FileReader();
+      fr.onload = function (e) {
```
jamesdools: this throws a linting error as `e` is defined in upper scope. rename it (and also would prefer arrow functions if possible)

src/index.js
```diff
@@ -58,22 +60,28 @@ class App extends React.Component {
...
-      console.log(JSON.parse(e.target.result));
-      self.setState({
-        transcriptData: JSON.parse(e.target.result),
-        // mediaUrl: tedTalkVideoUrl
-      });
-    };
-    fr.readAsText(file);
+  handleChangeLoadTranscriptJson(e) {
+    const files = e.target.files;
+    if (this.state.mediaUrl === null) {
```
jamesdools: try and separate these blocks with a line break maybe

src/index.js
```diff
@@ -117,6 +125,9 @@ class App extends React.Component {
      this.setState({ analyticsEvents: [ ...this.state.analyticsEvents, event ] });
    }
 
+   handleChangeTranscriptName = (value) => {
+     this.setState({ fileName: value });
+   }
```
jamesdools: line break after this

src/index.js
```diff
@@ -18,7 +18,8 @@ class App extends React.Component {
       mediaUrl: null,
       isTextEditable: true,
       sttType: 'bbckaldi',
-      analyticsEvents: []
+      analyticsEvents: [],
+      fileName: 'Kate Darling Ted Talk'
```
jamesdools: Not directly related to this PR - but I was thinking this demo app/wrapper around the component might actually make more sense in a different folder (ie. it shouldn't be in `/src/`, but a parallel folder called `demo` or something).

Similarly, `sample-data` might not belong in `/src`.

src/index.js
```diff
@@ -154,6 +153,17 @@ class App extends React.Component {
...
+         <br/>
+         <label>open Transcript Json</label>
+         <SttTypeSelect
+           name={ 'sttType' }
+           value={ this.state.sttType }
+           handleChange={ this.handleSttTypeChange }
+         />
+         <input
+           type="file"
+           onChange={ e => this.handleChangeLoadTranscriptJson(e) }
```
jamesdools: does this need the `e` passed in? Or does ` onChange={ this.handleChangeLoadTranscriptJson }` work?

src/lib/TranscriptEditor/TimedTextEditor/index.js
```diff
@@ -145,29 +145,48 @@ class TimedTextEditor extends React.Component {
   }
 
   localSave = () => {
-    const mediaUrl = this.props.mediaUrl;
+    let mediaUrlName = this.props.mediaUrl;
+    // if using local media instead of using random blob name
+    // that makes it impossible to retrieve from on page refresh
+    // use file name
+    if (this.props.mediaUrl.includes('blob')) {
```
jamesdools: Seems quite arbitrary, is this going to stay as the convention for now? We could have a more solid constant defined at the top of the file so you're not checking 'blob' in so many places?

src/lib/TranscriptEditor/TimedTextEditor/index.js
```diff
@@ -145,29 +145,48 @@ class TimedTextEditor extends React.Component {
...
     return false;
   }
 
   loadLocalSavedData(mediaUrl) {
-    const data = JSON.parse(localStorage.getItem(`draftJs-${ mediaUrl }`));
+    let mediaUrlName = mediaUrl;
+    if (mediaUrl.includes('blob')) {
+      mediaUrlName = this.props.fileName;
+    }
+    const data = JSON.parse(localStorage.getItem(`draftJs-${ mediaUrlName }`));
```
jamesdools: spacing out these blocks would be cleaner (to my eyes at least!)

src/index.js
pietrop: ok, to try this out I moved `sample-data` outside of `src` and I get this error:

```
Module not found: You attempted to import ../sample-data/KateDarling_2018S-bbc-kaldi.json which falls outside of the project src/ directory. Relative imports outside of src/ are not supported.
```

See `create-react-app` docs for more details 

https://facebook.github.io/create-react-app/docs/folder-structure

Also see our [README#system-architecture](https://github.com/bbc/react-transcript-editor#system-architecture.

src/lib/TranscriptEditor/TimedTextEditor/index.js
pietrop: I am going to leave it for now, coz it would need re-thinking how it is communicating with the parent component, eg removing the use of `ref`, but suggestions for that refactor welcome.

src/index.js
pietrop: Tried removing the `e` and it crashes.

src/index.js
pietrop: good catch, with the arrow function can also remove the `self = this` at the top.

PR #94, opened 2019-02-08
Speechmatics adapter

murezzda:
This is related to #38.

**Describe what the PR does**    
Adds a STT adapter for speechmatics.

Tests are not implemented yet.

**Additional context**    
Added sample transcript from speechmatics of the Demo TED Talk.


murezzda:
PR is ready for review.

Added the following functionality:

-Speechmatics transcripts can now be loaded.
-Speechmatics speakers are added to paragraphs.

pietrop:
Thanks @murezzda , I've a question in Speechmatics, speaker diarization is always provided or is it optional? (eg does it have extra cost?)

was looking at their [API docs](https://app.speechmatics.com/api-details) but it's unclear.

pietrop:
Looks good, as agreed, you can leave grouping the text by speaker segments for another PR.

closes https://github.com/bbc/react-transcript-editor/issues/38

murezzda:
Hi @pietrop 

> Thanks @murezzda , I've a question in Speechmatics, speaker diarization is always provided or is it optional? (eg does it have extra cost?)
> 
> was looking at their [API docs](https://app.speechmatics.com/api-details) but it's unclear.

Diarization is always provided by their cloud service by default and it does not cost extra. But it can be disabled by an argument. Maybe we should check for this in the next pr, good point.

pietrop:
Yes I think it might be good to make it optional. ( in separate PR)

Eg In BBC Kaldi adapter because speaker diarization is a separate optional attribute, what the adapter does is to check if the segments have been provided or not.
If they have it uses it. If not it has a fallback to do the parsing without.

I reckon this might be a good approach as could make the component adapter a bit more flexible. Eg someone wants to use it in a context where they already have run transcriptions through STT but didn’t think of getting the speaker diarization info etc..

Shizen-no-ko:
Hello, I would like to ask about the speechmatics adapter. I believe the json that speechmatics is returning these days might well be different to what this adapter was written for? It is erroring straight away on tmpWords = curatePunctuation(speechmaticsJson.words); because speechmatics json does not return a "words" key. I am presuming the best way would be for me to pre-filter the json I am getting back with its array of "alternatives", and create a "word" key?

Shizen-no-ko:
Hello, 
I have created a function to revert speechmatics V2 SaaS Json into V1 Json. I have had to use the npm module rather than the repo, as many components/modules are deprecated. However, if I leave punctuation marks in my conversion then the speechmatics adapter errors-out on the regEx test. But if I change my function to remove all punctuation marks, then your adapter works. Would you have any idea why? I'm stuck.

const curatePunctuation = (words) => {
  const curatedWords = [];
  words.forEach((word) => {
    if (/[.?!]/.test(word.name)) {  <------- Here

Also, this index.js file is presumably created by the minified code in the npm module. Is there any way of accessing this, to make logs/edits?

Sorry for the newbie questions.

Issue #97, opened 2019-02-13
Put focus on text when pause via shortcuts


murezzda:
**Is your feature request related to a problem? Please describe.**
If the player is stopped via keyboard shortcut, the focus will not be on the text and it is not possible to quickly edit the script only via shortcuts.

**Describe the solution you'd like**
Upon pausing the player via shortcuts, the curser will be set to the word corresponding to the current timestamp. This allows for a quick correction of false transcribed words.

**Describe alternatives you've considered**
The alternative is to rely on the autoscroll feature and manually move the focus after pausing.

**Additional context**
It would be great if the player was usable only via keyboard, as this would speed up the correction of the transcript.

pietrop:
Sounds great, thanks for raising this @murezzda !

Good to keep an eye out on how this will also work with the [`scroll-sync`](https://github.com/bbc/react-transcript-editor/blob/master/docs/qa/3-settings.md#item-to-test-35-settings---scroll-sync---on) feature (that when enabled in settings, keeps the current text in view)

PR welcome, if you want to give this a go :)

murezzda:
@pietrop working on a PR, should be ready next week.

pietrop:
explored in PR #102 but parked for now.

Issue #50, opened 2018-12-14
Suggested Layout change 


pietrop:
Just a suggestion on layout change to maximise space of text. Pending discussion. And low priority.

Current layout as of PR https://github.com/bbc/react-transcript-editor/pull/46

![controls](https://user-images.githubusercontent.com/4661975/50004506-3e01fa00-ff9f-11e8-8314-3b8983e3f247.png)

The thinking here is that there is a fair bit of space between the player controls and the transcript title, that is essentially due to the height of the video preview element. And it's taking up space that could be used to increase the height of the TimedTextEditor. 

So the suggestion to be discussed, would be to move the video player next to the TimedTexEditor, and below the player controls. As well as the progress bar over the player controls. - see balsamiq sketch below.

![controls layout change](https://user-images.githubusercontent.com/4661975/50004491-36daec00-ff9f-11e8-9648-27cc17f04bed.png)

On a 15 inch Mac this might not make much of a difference, but on smaller size laptops or tablets(?) this change could be a bit more significant.

pietrop:
Just to add to the mix of consideration, the current audio only combination looks a bit off 

![screen shot 2019-01-17 at 14 47 02](https://user-images.githubusercontent.com/4661975/51326378-e6a3ed80-1a66-11e9-83a2-415b2462193d.png)

pietrop:
ah, I see what you mean now.

pietrop:
done some more work on branch to explore this further https://github.com/bbc/react-transcript-editor/pull/70

pietrop:
current PR for this is https://github.com/bbc/react-transcript-editor/pull/111

pietrop:
addressed in https://github.com/bbc/react-transcript-editor/pull/115

Issue #49, opened 2018-12-14
Mobile Responsive 


pietrop:
It be great if the desktop layout could be mobile responsive

![controls](https://user-images.githubusercontent.com/4661975/50004220-676e5600-ff9e-11e8-801c-66eeb1323fc1.png)

Here's one possible view for mobile. with the 4 main blocks stacked in this order
- Video Preview
- Player Controls
- Progress bar
- TimedTextEditor

![mobile](https://user-images.githubusercontent.com/4661975/50004221-676e5600-ff9e-11e8-98e9-17f281323452.png)

- [ ] To be decided, is whether to have video preview element as fixed, and a toggle in settings to hide display of preview to gain more space for text on small screen.

pietrop:
addressed in https://github.com/bbc/react-transcript-editor/pull/115

Issue #30, opened 2018-11-26
A way of preserving or restoring time-codes while or after editing the text


pietrop:

pietrop:
I've tried the following experiment

- [x]  Take Chapter One of Moby Dick [](https://librivox.org/moby-dick-by-herman-melville/)
   - [Audio from LibriVox](http://ia802604.us.archive.org/32/items/moby_dick_librivox/mobydick_000_melville.mp3)
  - [Text from project Gutemberg](http://www.gutenberg.org/ebooks/2701) - [Plain text version - only chapter 1](http://www.gutenberg.org/files/2701/2701-0.txt)

- [x] Tun the LibriVox audio through BBC Kaldi (You could use Gentle or some other STT service tho)

- [x] Run through [stt-align-node](https://github.com/bbc/stt-align-node) (open sourcing soon 🤞) (aligning algo via @chrisbaume )- This aligns a known accurate transcript to the output of a speech-to-text service. Transposing the times of the words to the accurate text.

- [x] Got [a re-aligned json](https://github.com/bbc/stt-align-node/blob/master/sample/output/modbydick-realigned.json) file for Moby Dick Chapter 1.

- [x] Modified the re-aligned json to be closer to kaldi by replaced `word` with `punct`. to be able to re-import into the TranscriptEditor. 

- [x] Imported into TranscriptEditor, and checked, start, middle, and end of piece, and seems like the alignment is spot on 🙌

![screen shot 2018-12-20 at 16 55 08](https://user-images.githubusercontent.com/4661975/50295431-4d3de780-0478-11e9-8eff-1ce28d09d14b.png)

pietrop:
So it be good to test this out more extensively but also to try out the "wave form comparison approach" eg [WebAligner](https://github.com/chrisbaume/webaligner) - similar to [Aeneas](https://github.com/readbeyond/aeneas/) but perhaps possible client side (?).

pietrop:
I've tried with the same audio and text from the [Ted Talk](https://www.ted.com/talks/kate_darling_why_we_have_an_emotional_connection_to_robots/transcript?language=en) used in the demo app.

- [x] Got the accurate transcription from the ted talk website [ted-talk-kate.txt](https://github.com/bbc/react-transcript-editor/files/2699848/ted-talk-kate.txt)
[...]

PR #102, opened 2019-02-20
Focus on pause

murezzda:
Related to issue #97

**Describe what the PR does**    
After pausing the player via the alt+k shortcut, the cursor is moved to the current word.

PR still needs some testing, but functionality wise works.

### DIFF
--- a/src/lib/TranscriptEditor/MediaPlayer/VideoPlayer.js
+++ b/src/lib/TranscriptEditor/MediaPlayer/VideoPlayer.js
@@ -13,6 +13,7 @@ class VideoPlayer extends React.Component {
         data-testid="media-player-id"
         onClick={ this.props.onClick }
         onLoadedData={ this.props.onLoadedDataGetDuration }
+        onPause={ this.props.videoHasPaused }
         ref={ this.props.videoRef }
       />
     );
@@ -24,7 +25,8 @@ VideoPlayer.propTypes = {
   onTimeUpdate: PropTypes.func,
   onClick: PropTypes.func,
   videoRef: PropTypes.object,
-  onLoadedDataGetDuration: PropTypes.func
+  onLoadedDataGetDuration: PropTypes.func,
+  videoHasPaused: PropTypes.func
 };
 
 export default VideoPlayer;
--- a/src/lib/TranscriptEditor/MediaPlayer/defaultHotKeys.js
+++ b/src/lib/TranscriptEditor/MediaPlayer/defaultHotKeys.js
@@ -4,7 +4,7 @@ function returnHotKeys(self) {
       priority: 1,
       handler: () => {
         self.togglePlayMedia();
-        if (!self.isPlaying()) {
+        if (!self.isPlaying() && self.props.focusOnPause) {
           self.props.setFocus();
         }
 
--- a/src/lib/TranscriptEditor/MediaPlayer/defaultHotKeys.js
+++ b/src/lib/TranscriptEditor/MediaPlayer/defaultHotKeys.js
@@ -4,6 +4,9 @@ function returnHotKeys(self) {
       priority: 1,
       handler: () => {
         self.togglePlayMedia();
+        if (!self.isPlaying()) {
+          self.props.setFocus();
+        }
 
         self.props.handleAnalyticsEvents({
           category: 'defaultHotKeys',
--- a/src/lib/TranscriptEditor/MediaPlayer/defaultHotKeys.js
+++ b/src/lib/TranscriptEditor/MediaPlayer/defaultHotKeys.js
@@ -4,9 +4,6 @@ function returnHotKeys(self) {
       priority: 1,
       handler: () => {
         self.togglePlayMedia();
-        if (!self.isPlaying() && self.props.focusOnPause) {
-          self.props.setFocus();
-        }
 
         self.props.handleAnalyticsEvents({
           category: 'defaultHotKeys',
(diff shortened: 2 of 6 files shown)


pietrop:
👋 @murezzda , thanks for this!
Tested the functionality, seems to be working fine, before I start reviewing the code, wanted to check if this PR is ready for review?

murezzda:
Hi @pietrop, I will have a final look at the PR tomorrow and give you notice as soon as I've checked it again.

pietrop:
I think you mentioned in `TimedTextEditor` `getCurrentFocus()` line `494` that

>  using convertToRaw here might be slowing down performance(?)

Which is something I might have noted elsewhere in the codebase, have you come across any alternatives? (otherwise we could roll with that for now)

murezzda:
Hi @pietrop 
I've checked the code again and it would be ready for review.

I have still the following points open, maybe you have an option/idea about it:

1. **Where to put cursor / whether to select word or not.**

- At the moment the cursor is placed at the end of the current word
- Alternatives could be to place it at the beginning of the current word or to pre-select the current word.
- Pre-selection would be a good option, but doesn't seem to be intuitive with the current layout, as the selected text has the same color as the text highlight from the player. Also it might lead to accidental deletion of words if people are not cautious.

2. **Performance and word selection lag**

- Regarding your question about the comment at line `494`, I've based the methon on the already present `getCurrentWord` method which also has this comment in order to be able to find it if an alternative was found. At the moment I do not have an alternative.
- The performance of the word selection seems to be wanting at times, as the video highlight tends to move ahead of the selection if the spoken words are short. In my opinion this is does not break the feature, but could be annoying to the user.

pietrop:
👋 @murezzda, thanks for the PR would you be able to add an option in settings to make this optional?
Let me know, I can help/show how that would work. 
With @jamesdools , we were thinking that it could be on by default, but with an option to turn it of?

I also check with @Laurian re-using `convertToRaw` and he said it should be ok from a performance point of view.

murezzda:
Hi @pietrop, making this feature optional makes also sense to me. I will check out the options code and will contact you if there are any questions.

If its not the problem of `convertToRaw ` I am going to look over the selection loop again. I have a suspicion that there could be a better solution than the current one.

Thanks for the input!

murezzda:
Hi @pietrop, I've added now the option of turnign this feature off in the settings and I was also able to fix the issue with the lagging selection. The side-effect of fixing this is that the cursor will also be moved if the user pauses the video via the video-controls, but in my opinion this is anyway the sensible thing to do.

Feel free to review the code, it should be ok now.

murezzda:
Setting focus whenever a pause is triggered is not compatible with the pause-while-typing. If both are active, the cursor will move to the position of the video player whenever someone types during the video. This will prevent the user from correcting the word.

pietrop:
ah, ok interesting, thanks for this, going to find some time have a look and running locally.

pietrop:
Thanks @murezzda 

**1. Where to put cursor / whether to select word or not.**

I think this seems fine 

> At the moment the cursor is placed at the end of the current word

We could also add a reminder in keyboard shortcuts that `alt`+`shift`+`→` select the next word

And that `alt`+`shift`+`←` select previous word.

These are defaults behaviour, supported by draftJS.


**2.Performance and word selection lag**
I didn't notice any lags so far.

**4. cursor in selection**
Seems like the cursor gets "trapped" in the current word selection when one is present.

in the gif below you can see how if a word is not selected eg end of sentence, it works fine, but that if a current word is selected then it would add the text inside the selection. Then selection will disappear when playing again.
bug-focus-on-pause

This is something that might look as simple technical glitch, but I can see how it might be confusing (and possibly mildly annoying?) for some users.

**5. Settings option**
Thanks for adding the optional toggle in settings.

**6. not compatible with pause while typing**

You mentioned

> Setting focus whenever a pause is triggered is not compatible with the pause-while-typing. If both are active, the cursor will move to the position of the video player whenever someone types during the video. This will prevent the user from correcting the word.

And I can confirm the problem.

bug-focus-on-pause-pause-while-typing


This makes me think, if we can't fix this, it should then not be on by default.
A workaround is that If it was off by default, then in the settingss, toggling this on, would turn pause while typing off to avoid conflict, and user can choose which one to use.

---

### TL;DR
One of our assumption, that we are about to user test, is whether pause while typing, is a key feature while correcting automated transcript.

After having played around with it a bit more. I am inclined to say that we should probably close and park this for now. We can do further research and exploration and a later stage if it comes up as a user need. But wouldn't want to fall into the trap of premature optimisation of the tool and feature creep. As we are aiming to spend a bit of time to consolidate the editor and refactor the code, to make it more modular and more robust. Giving priority to https://github.com/bbc/react-transcript-editor/issues/50 + https://github.com/bbc/react-transcript-editor/issues/49 + https://github.com/bbc/react-transcript-editor/issues/30

murezzda:
Hi @pietrop 

Thanks for the in-depth look at the PR!

I have a few additional comments, but generally I agree with your assessment.

**Performance and word selection lag vs non-compatibility with pause-on-typing**
The performance issues I've described previously where from the version before the last commit in this PR. In that version, the moving of the cursor was done inside the hotkey evaluation code. This ensured that focus was only set when a pause via hotkey happened. I changed this for two reasons: First, I thought it too confusing to restrict this functionality to hotkeys only. The second reason was that when this was done just after invoking `self.togglePlayMedia();`, the pausing of the media player and the moving of the cursor was not synchronous.

The current version moves the cursor in the `onPause` event in the Video Player. This ensures that the cursor is only moved _after_ the player has stopped, but introduces the clash with the pause while typing feature as it is no longer possible to track the origin of the pause event.

I guess one could hack this to work by adding more state to track the origin of the event, but I think that would neither be a good nor a maintainable solution...

**Cursor in Selection**
Thanks for catching this. I will look at this -if- I will work on this feature in the future.

**Verdict**
I agree that we should park this feature for the moment, as there seem to be some issues which are not easily resolved. I will post an update if there are further developments which would solve these issues.

pietrop:
Awesome, will close this for now, but we can revisit at later stage, thank you @murezzda

Issue #98, opened 2019-02-13
Moving cursor in text puts focus on timestamp and speaker labels


murezzda:
**Is your feature request related to a problem? Please describe.**
When the cursor is moved via arrow keys in the editor-box, it is not possible to quickly move between paragraphs, as the cursor will first move to the timestamp and the speaker label, both which cannot be edited. This hinders quick movement inside the text via keyboard.

**Describe the solution you'd like**
Timestamps and speaker labels will no longer come into focus if the curser is moved inside the text area.

**Additional context**
This feature would help make the editor better usable only via keyboard.

murezzda:
Added PR which mostly solves the issue (see PR description).

PR #103, opened 2019-02-20
Moving cursor no longer puts focus on timestamp and speaker labels

murezzda:
This PR addresses #98.

**Describe what the PR does**    
Added `contentEditable={ false }` to speaker and timecode elements. This prevents the cursor from entering the timecode element. The cursor still selects speaker element when moving cursor out of the last character of a block, but not when switching between blocks directly.

Cursor still selects speaker element when moving cursor out of the last character of a block. This needs to be fixed before review.

### DIFF
--- a/src/lib/TranscriptEditor/TimedTextEditor/SpeakerLabel.js
+++ b/src/lib/TranscriptEditor/TimedTextEditor/SpeakerLabel.js
@@ -9,8 +9,7 @@ import style from './WrapperBlock.module.css';
 class SpeakerLabel extends PureComponent {
   render() {
     return (
-      <span className={ style.speaker }
-        contentEditable={ false }>
+      <span className={ style.speaker }>
         <span
           className={ style.EditLabel }
           onClick={ this.props.handleOnClickEdit }>
--- a/src/lib/TranscriptEditor/TimedTextEditor/SpeakerLabel.js
+++ b/src/lib/TranscriptEditor/TimedTextEditor/SpeakerLabel.js
@@ -9,7 +9,8 @@ import style from './WrapperBlock.module.css';
 class SpeakerLabel extends PureComponent {
   render() {
     return (
-      <span className={ style.speaker }>
+      <span className={ style.speaker }
+        contentEditable={ false }>
         <span
           className={ style.EditLabel }
           onClick={ this.props.handleOnClickEdit }>
--- a/src/lib/TranscriptEditor/TimedTextEditor/WrapperBlock.js
+++ b/src/lib/TranscriptEditor/TimedTextEditor/WrapperBlock.js
@@ -93,16 +93,12 @@ class WrapperBlock extends React.Component {
       handleOnClickEdit={ this.handleOnClickEdit }
     />;
 
-    const timecodeElement = <span
-      contentEditable={ false }
-      className={ style.time }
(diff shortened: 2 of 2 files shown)


pietrop:
That's great thanks @murezzda 

## Cursor colour 

I found it difficult to follow where the cursor is when scrolling down the text, so thought might be a good idea to give it a different colour, eg [red or blue](https://css-tricks.com/almanac/properties/c/caret-color/)?

eg

```js
[contenteditable] {
  caret-color: red;
}
```

But perhaps scoped only to the Timed Text Editor (otherwise it will effect every other conte editable on the page on the parent application)

In the body of the text it works fine 🎉 

draftjs

## Issue when reaching top of page
There seems to be an edge case when reaching the top of the page and wanting to navigate back down.

When reached top of page if click arrow up cursor moves to the sidebar scroll rather then the cursor. 
if that makes sense? see gif below.

To reproduce
- Click in the timed text editor
- Arrow up  ↑ all the way to the end of the timed text editor
- Arrow down ↓
- You should notice that the cursor is disappeared and the arrow now controls the "sidebar" for the timed text editor.

draftjs3


As a side note, this does not seem to happen in the current master (try in [demo](https://bbc.github.io/react-transcript-editor/) - load demo)

pietrop:
In `WrapperBlock.js` it seems like a fix might be to add the `contentEditable={ false }` to the div that contains both the `speakerElement` and the `timecodeElement` only and with that change it seems to work fine 🎉 

``` line 104 
<div className={ style.WrapperBlock }>
        <div className={ style.markers } contentEditable={ false }>
          {this.props.blockProps.showSpeakers ? speakerElement : ''}
          {this.props.blockProps.showTimecodes ? timecodeElement : ''}
        </div>
        <div className={ style.text }>
          <EditorBlock { ...this.props } />
        </div>
      </div>
```

src/lib/TranscriptEditor/TimedTextEditor/SpeakerLabel.js
```diff
@@ -9,7 +9,8 @@ import style from './WrapperBlock.module.css';
 class SpeakerLabel extends PureComponent {
   render() {
     return (
-      <span className={ style.speaker }>
+      <span className={ style.speaker }
+        contentEditable={ false }>
```
pietrop:
```suggestion
       >
```

src/lib/TranscriptEditor/TimedTextEditor/WrapperBlock.js
```diff
@@ -93,7 +93,12 @@ class WrapperBlock extends React.Component {
       handleOnClickEdit={ this.handleOnClickEdit }
     />;
 
-    const timecodeElement = <span className={ style.time } onClick={ this.handleTimecodeClick }>{shortTimecode(startTimecode)}</span>;
+    const timecodeElement = <span
+      contentEditable={ false }
```
pietrop:
```suggestion
     
```

murezzda:
Hi @pietrop, this does solve this problem indeed. Thanks a lot, I've pushed an updated version already.

pietrop:
Awesome thanks @murezzda !

murezzda:
Hi @pietrop 

I've just wanted to follow up on your comment about the cursor color. I think a more prominent cursor would be a great idea.

Issue #108, opened 2019-03-10
Clarification on how to create an Adapter for AWS Transcribe


Gribbs:
I've forked the repo and I'm trying to create a new Adapter for AWS Transcribe formats but I'm having trouble understanding the recommended steps here https://github.com/bbc/react-transcript-editor/blob/master/docs/guides/adapters.md

The recommended steps are:

 1. Create a folder with the name of the STT service - eg speechmatics
 2. add a adapters/${sttServiceName}/sample folder
 3.  add a sample json file from the STT service in this last folder - this will be useful for testing. 
 4. Name it ${name of the stt service}.sample.json
 5. add option in adapters/index.js

I can see an adapters folder in the project here src/lib/Util/adapters .

Is this where we should begin with step 1? I'm assuming yes, but just want to make sure this is right.

pietrop:
Hi @Gribbs,
Thanks for reaching out and the PR.
Yes, that’s the right location, will add a note to the docs make it explicit.

chrishutchinson:
Hey @Gribbs, I had started throwing together an AWS adapter too, but I haven't got round to finishing it up yet.

What sort of progress have you made so far? I'm keen to see this, so am happy to help if it's useful.

Gribbs:
hey @chrishutchinson, I've completed the implementation (might be a bit untidy in there with some cut and pasted files in the sample folder though) on my fork and seems to work ok.

I haven't implemented speakers yet though. I also was a bit confused with how to use the generic generateEntitiesRanges() function on the Adapters guide:
**4. And use the helper function generateEntitiesRanges to add the entityRanges to each block. - see above**
The guidance recommends using it like generateEntitiesRanges(paragraph.words, 'text')

However, for Transcribe the "text" lives on an Object in each word like:

```
const generateEntitiesRanges = (words, wordAttributename) => {
 const text = word.alternatives[0].content //  <--- I can't use wordAttributename param here
  // rest of code
}
```
I figured I'd either have to remap each word or use my own implementation of generateEntitiesRanges() instead. I chose the latter. I'm not sure if I should be doing it different though?

chrishutchinson:
Sounds good @Gribbs, this is about as far as I got with my prototype. I ended up with similar questions but didn't make any progress resolving them.

[...]

PR #88, opened 2019-01-23
Fix 87 initialising transcript media - component interface

pietrop:
**Is your Pull Request request related to another [issue](https://github.com/bbc/react-transcript-editor/issues) in this repository ?**      
_If so please link to other issues and PRs as appropriate_

https://github.com/bbc/react-transcript-editor/issues/87

**Describe what the PR does**    
_A clear and concise description of what the PR does. Feel free to use bulletpoints and checkboxes if needed [...]_

As explained in the QA section

> The component at a minimum needs two params to function properly and to be able to save to local storage. Transcript as `transcriptData` (+ `sttJsonType`) and media as`mediaUrl`.

> We'd normally expect both to be provided at the same time, but there might be edge case where the component is initialized with the first one, and then subsequently receives the second one (or the other way around).

This PR addresses this issue, making it possible to initialise the component with either both, or with one or the other first subsequently followed by the other.

**State whether the PR is ready for review or whether it needs extra work**    
_If you are still working on it and just setting it up for later review, or if it's ready to be reviewed for merging_

almost ready for review

**Additional context**    
_Add any other context or screenshots about the PR._

NA

Issue #87, opened 2019-01-23
If initialising component with Transcript first and then Media local storage save/retrieve does not work


pietrop:
**Describe the bug**
_A clear and concise description of what the bug is._

The component at a minimum needs two params to function properly and to be able to save to local storage:
- media `mediaUrl`
- transcript `transcriptData` (+ type)

as background saving to local storage uses the media url as a key.

- [X] Providing both at the same time <-- works

- [X] Providing first media then transcript <-- works

- [ ] First transcript then media <-- breaks

**To Reproduce**
1. Go to https://bbc.github.io/react-transcript-editor/
2. Click on 'choose file' next to 'open Transcript Json'  
3. Click on 'choose file' Load Local Media'  
4. Click on text and edit, (add more then 5 char)  
5. refresh page
6. repeat 1-5 and see the text added in step 4 is present
4. if not, then this is the error

**Expected behavior**
_A clear and concise description of what you expected to happen._

It would be preferable if the component could handle local storage save even if transcript is provided before the media is passed to it.

**Additional context**
_Add any other context about the problem here._

Possible solution if Media not defined then transcript should wait to load until media is. Binding TranscriptEditor component to only work if both are present.

PR #110, opened 2019-03-13
Stt adapter awstranscribe

Gribbs:
  
Yes. Original issue https://github.com/bbc/react-transcript-editor/issues/108

**Describe what the PR does**    

The PR add an Adapter to support Amazon Transcribe https://aws.amazon.com/transcribe/

Ready to review for merging.

**Additional context**    
The PR doesn't cater for Speakers at this point (although the Transcribe service does support Speakers).
I haven't done an example-usage.js file in there or or .test.js but I do have an example Transcribe json file in there.

Steps to test the changes:

yarn start or npm start. I have added the option 'Amazon Transcribe' in src/select-stt-json-type.js


pietrop:
Awesome, thanks @Gribbs ,
I'll find some time to have a look at this by the of the week.

for now I'd say: 
- ok to leave speaker identification for later stage. - eg as separate PR.
- You can probably consider`example-usage.js` optional, as that can be added later. 
- But would you be able to add the`.test.js` test file so that the Travis CI doesn't fail? _Let me know if you need any help with that._

---

cc @chrishutchinson

package.json
```diff
@@ -37,7 +37,6 @@
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "babel-preset-minify": "^0.5.0",
     "enzyme": "^3.7.0",
```
pietrop: did you remove eslint?

package.json
```diff
@@ -37,7 +37,6 @@
...
-    "eslint": "5.6.0",
```
pietrop: did you remove eslint?

src/lib/Util/adapters/amazon-transcribe/index.js
```diff
@@ -0,0 +1,112 @@
...
+  });
+
+  return results;
+};
+
+const amazonTranscribeToDraft = (amazonTranscribeJson) => {
+  const results = [];
+  let tmpWords;
+
+  // BBC Octo Labs API Response wraps Kaldi response around retval,
```
pietrop:
```suggestion
  // BBC Octo Labs API Response wraps AWS response around retval,
```

chrishutchinson:
Hey @Gribbs - I'll take a skim through the implementation too, but having run it locally with a test video / transcript from AWS, it seems to be working really well, nice one!

I don't seem to be able to download the resultant edited transcript, but I'm guessing that's either a bug with Firefox, or something else, not specific to this code change. cc @pietrop are you aware of anything? I'll do further digging if not and file a separate issue.

pietrop:
Thanks @chrishutchinson ,

haven't had a chance to try it in the PR yet but I did a quick test in the demo app  https://bbc.github.io/react-transcript-editor (`load demo`)

I am able to download draftJs and plain text of current demo  in Chrome but not in firefox.

Firefox console says

```
TypeError: t.refs.timedTextEditor is undefined
```

which is something to do with `getEditorContent` in the example/demo app. 

Which at the moment is implemented this way

```js
  getEditorContent = (exportFormat) => {
    return this.refs.timedTextEditor.getEditorContent(exportFormat);
  }
```

But could be refactored as a callback? and not user `refs` - [which is not recommended by React](https://reactjs.org/docs/refs-and-the-dom.html) and see if it also fixes it in Firefox? -  this could be a separate PR.

Do you get the same error? (I am aware the demo is a different version number compared to master, and compared to the PR)

package.json
Gribbs: I did remove eslint in the package.json . I was getting a CRA error similar (if not identical) to this issue https://github.com/facebook/create-react-app/issues/5247 so I removed it from package.json and nuked my node_modules and did an npm install and all was well after that. My apologies, I forgot I did that

Gribbs: I'll add a .test.js today too

Gribbs: That error that made me override the delete the eslint from package.json has resurfaced for some reason:

...........................
yarn start
yarn run v1.13.0
$ react-scripts start

There might be a problem with the project dependency tree.
It is likely not a bug in Create React App, but something you need to fix locally.

The react-scripts package provided by Create React App requires a dependency:

  "eslint": "5.12.0"

Don't try to install it manually: your package manager does it automatically.
However, a different version of eslint was detected higher up in the tree:

  /Users/gjoel/Documents/reactapps/react-transcript-editor/node_modules/eslint (version: 5.6.0)

Manually installing incompatible versions is known to cause hard-to-debug issues.

If you would prefer to ignore this check, add SKIP_PREFLIGHT_CHECK=true to an .env file in your project.
That will permanently disable this message but you might encounter other issues.

To fix the dependency tree, try following the steps below in the exact order:

  1. Delete package-lock.json (not package.json!) and/or yarn.lock in your project folder.
  2. Delete node_modules in your project folder.
  3. Remove "eslint" from dependencies and/or devDependencies in the package.json file in your project folder.
  4. Run npm install or yarn, depending on the package manager you use.

In most cases, this should be enough to fix the problem.
If this has not helped, there are a few other things you can try:

  5. If you used npm, install yarn (http://yarnpkg.com/) and repeat the above steps with it instead.
     This may help because npm has known issues with package hoisting which may get resolved in future versions.

  6. Check if /Users/gjoel/Documents/reactapps/react-transcript-editor/node_modules/eslint is outside your project directory.
     For example, you might have accidentally installed something in your home folder.

  7. Try running npm ls eslint in your project folder.
     This will tell you which other package (apart from the expected react-scripts) installed eslint.

If nothing else helps, add SKIP_PREFLIGHT_CHECK=true to an .env file in your project.
That would permanently disable this preflight check in case you want to proceed anyway.

P.S. We know this message is long but please read the steps above :-) We hope you find them helpful!

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
.........................................

Gribbs: I'll add .env temporarily so I can at least test the .test.js file

Gribbs: ok now added src/lib/Util/adapters/amazon-transcribe/index.test.js and pushed the changes. I did notice the test doesn't have a status of PASS but a status of RUNS and it doesn't exit. I'm not sure why that is though.
Do you need anything else from me for this PR @pietrop ?

pietrop:
Thanks for this PR @Gribbs!

**`generateEntitiesRanges`**
Glad it worked out to use the shared helper function

**Test**
The amazon transcribe test fails, because the test file `amazonTranscribe.sample.js` needs  [this setup step](https://github.com/bbc/react-transcript-editor/blob/master/docs/guides/adapters.md#tests).

In ` src/lib/Util/adapters/amazon-transcribe/sample/amazonTranscribe.sample.js` all the entity ranges are as 

```js
"entityRanges": [{
		"start": 13.03,
		"end": 13.22,
		"confidence": 1,
		"text": "There",
		"offset": 0,
		"length": 5,
		"key": "ico58i6"
	}, {
```

and because the key is randomly generated it cannot be deterministically tested by jest.
So instead the attribute should be just expecting to be a string

```js
"key": expect.any(String)
```

As explained in the guide, you could do a find and replace in your code editor, if it supports regex for `"key": "([a-zA-Z0-9]*)"` and replace with the above. if that makes sense?

[you should be able to see detailed Travis CI log of the test failing here](https://travis-ci.org/bbc/react-transcript-editor/builds/506115982?utm_source=github_status&utm_medium=notification)

Your tests should then pass.

**Console errors**

I get this error in the console when adding a AWS transcript

```
index.js:1446 Warning: Received NaN for the `data-start` attribute. If this is expected, cast the value to a string.
    in span (at Word.js:36)
```

as a quick fix might be enough to do some type casting and add a `.toString()` to `src/lib/TranscriptEditor/TimedTextEditor/Word.js` at line 37-38

```js
  <span
        data-start={ data.start.toString() }
        data-end={ data.end.toString() }

```

**Adding transcript first and then media - error**
If i test by adding adding media first and then transcript, it works fine but wen adding transcript first and then media it crashes and gives some errors

```
Uncaught TypeError: Cannot read property 'isPresentInLocalStorage' of null
    at TranscriptEditor._this.ifPresentRetrieveTranscriptFromLocalStorage 
```
Altho to be fair, might not be specific of this branch - just tested on master and got the same.
Altho it works fine in [current demo page](https://bbc.github.io/react-transcript-editor/) - which is a few versions behind, so might need to do some investigating.

I think it might be connected with issue https://github.com/bbc/react-transcript-editor/issues/87 and PR  https://github.com/bbc/react-transcript-editor/pull/88 and somehow most likely the bug might have been re-introduced since.  

But since it's effecting master as well, I wouldn't necessary aim to figure this out as part of this PR. can be a separate issue/PR.

**example usage**
You could add this in example usage 

```js
import amazonTranscribeToDraft from'./index';
import amazonTranscribeTedTalkTranscript from './sample/autoEdit2TedTalkTranscript.sample.json';

console.log(amazonTranscribeToDraft(amazonTranscribeTedTalkTranscript));
```

**eslint**
It be good to re-introduce ESLint in the dev dependencies. did you followed the instructions you got with the error that you copy and pasted up there? do you have a copy of ESLint installed globally? 

---

### TL;DR:
would you be able to: 
- [ ] fix the test
- [ ] add the example usage
- [ ] restore the the linting

The other errors, since they are across the master branch can look into it with @jamesdools

Gribbs:
Thanks @pietrop pietrop. Ive added the eslint back in to dev dependencies.

I also was getting sporadic "Received NaN for the `data-start` attribute" too and couldn't work out what was causing it. I did notice the json returned from amazon transcribe provides the numbers wrapped in quotes like:

`  "start_time": "13.22"`

  compared to bbck-kaldi for example, which looks like:

`  "start": 13.44`

  and thought this might've been causing it originally even though I was casting to a number. I've had to do some awkward processing with punctuation start and end times since there is no start and end time provided by Transcribe. I've now added an addition .toFixed() method so my decimal places don't go on too long:

`  start: /punctuation/.test(currentWord.type) ? (parseFloat(previousWord.end_time) + 0.05).toFixed(2) : parseFloat(currentWord.start_time)`

  For now I'm not getting the error.

  For the testing, I've added the `"key": expect.any(String)` to the amazonTranscribe.sample.js but I still can't get the test to pass. I've tried to debug it but I can't work out how to get any visibility into what is making the test fail (or run forever). I've tried:

`  yarn test index.test.js`

  and

`  yarn test-ci`

  and

`  yarn test-ci --verbose`

  But nothing seems to give me an indication what the problem is. If you can help with that it would be appreciated

pietrop:
Thank for the changes @Gribbs , 

**eslint**
ah, my bad, I think you were right, `eslint` doesn't need to be installed in create react app as apparently it's there under the hood (cc @jamesdools for some of the other PRs eg https://github.com/bbc/react-transcript-editor/pull/115)

I had missed that in the top part error message - [noticed it in the travis error message](https://travis-ci.org/bbc/react-transcript-editor/builds/506970779?utm_source=github_status&utm_medium=notification)

```
There might be a problem with the project dependency tree.
It is likely not a bug in Create React App, but something you need to fix locally.
The react-scripts package provided by Create React App requires a dependency:
  "eslint": "5.12.0"
Don't try to install it manually: your package manager does it automatically.
```

**tests** 
No worries, I can have a look into it and see what might be the problem. Then I reckon it'd be good to merge to master.

Gribbs:
Alright sounds good! Thanks again for your help!

pietrop:
I noticed a bug, seems like punctuation such as commas is added is as it's own word?

It might be more straightforward to append it to the previous word as it doesn't come with start and end time informations.

eg

```json
{
        "alternatives": [
          {
            "confidence": null,
            "content": ","
          }
        ],
        "type": "punctuation"
      },
```

if double clicking the punctuation, eg `,` it crashes, as time is not associated to it?
```
index.js:79 Uncaught TypeError: Failed to set the 'currentTime' property on 'HTMLMediaElement': The provided double value is non-finite.
```

and just to confirm and I can see it get renders as with start and end time as `NaN`

```html
<span data-start="NaN" data-end="NaN" data-confidence="high" data-prev-times="" data-entity-key="c0a6o9g" class="Word">
	<span data-offset-key="64k16-22-0"><span data-text="true">,</span>
</span>
```

What could be an easy fix for this @Gribbs ?

Gribbs:
I thought I’d tested that. Maybe my last change with toFixed() has caused it? I was accounting for this by using the previous word end-time as the the next punctuation word start time + a tiny amount of time.

Gribbs:
I’ll take another look. I’ve just left the house so will look at it again this evening

pietrop:
ok no worries, with the change below, commas are also included into having start and end time

```js
const groupWordsInParagraphs = (words) => {
  const results = [];
  let paragraph = {
    words: [],
    text: []
  };
  words.forEach((word, index) => {
    // if word type is punctuation
    const content = word.alternatives[0].content;
    let previousWord = {};
    if (word.type === 'punctuation' && /[.?!]/.test(content)) {
      previousWord = words[index - 1]; //assuming here the very first word is never punctuation
      paragraph.words.push(normalizedWord(word, previousWord));
      paragraph.text.push(content);
      results.push(paragraph);
      // reset paragraph
      paragraph = {
        words: [],
        text: []
      };
    } else if (word.type === 'punctuation' && /[,?!]/.test(content)) {
      previousWord = words[index - 1]; //assuming here the very first word is never punctuation
      paragraph.words.push(normalizedWord(word, previousWord));
      paragraph.text.push(content);
    } else {
      paragraph.words.push(normalizedWord(word, previousWord));
      paragraph.text.push(content);
    }
  });

  return results;
};
```

altho it looks a little off, coz treating them like words there are then extra spaces before commas and full stops

Screen Shot 2019-03-15 at 22 04 10

so there might be a better way to do it, and append punctuation to the previous word.

pietrop:
I am going to have closer look next week

Gribbs:
I think your right. The spaces around the punctuation have been annoying me too. I’ll work on appending it to the previous word. That should simplify the code a bit too

Gribbs:
Hi @pietrop I've completed the work to append punctuation items to the previous word which seems to work well. I've pushed the changes to this branch. I've also added some additional tests in index.test.js file so it should pass ok. I'm skipping the original problematic test for now with describe.skip until I can work out what the issue with that test was.

Gribbs:
oh! I see this pull request has now been closed. Would you like me to put it on separate pull request?

pietrop:
Hi @Gribbs, yes, was reviewing it in `stt-adapter-awstranscribe-review`but since you made new changes feel free to setup a PR from your fork against that branch if that works for you?

Gribbs:
> Hi @Gribbs, yes, was reviewing it in branch `stt-adapter-awstranscribe-review`but since you made new changes feel free to setup a PR from your fork against that branch. if that works for you?

I've created PR https://github.com/bbc/react-transcript-editor/pull/116 for this

PR #70, opened 2019-01-04
Layout change

pietrop:
**Is your Pull Request request related to another [issue](https://github.com/bbc/react-transcript-editor/issues) in this repository ?**      

Suggested Layout change https://github.com/bbc/react-transcript-editor/issues/50

TL;DR: 
- Increasing the height of the space for text
- allowing to have video preview both resizable and toggle show/hide option.

**Describe what the PR does**    

- changed layout
- moved` VideoPlayer` preview into `TranscriptEditor` next to `TimedTextEditor`
- added optional transcript title (non editable) with css logic to truncate if text is too long
- added options in settings to hide preview
- added option in settings to make video preview larger or smaller within custom range  

**State whether the PR is ready for review or whether it needs extra work**    

Still doing some tweaks, and cleaning up, should be ready for review soon.

Outstanding
- [ ] Some of the logic across components might need refactoring and cleaning up - simplifying
- [ ] The controls with all the playerControls btns could do with some spacing, to group by logic eg adding a space after fast forward, and after playback speed, before volume, and before title. (?)
- [ ] Update QA doc with new functionalities (QA PR needs to merged into master before this one)
- [ ] check that analytics events are still triggered (from QA doc list) - there might be one or two that are broken 

**Additional context**    

screen shot 2019-01-04 at 12 00 04

PR #111, opened 2019-03-14
Layout changes accessibility

ivanji:
https://github.com/bbc/react-transcript-editor/issues/50

**Describe what the PR does**    
Continuing on work done by @pietrop  on #70 

Ready for review

**Additional context**    
editor-layout-changes


pietrop:
Thanks for this @ivanji , we are having a closer look with @jamesdools tomorrow. 

I reckon we could hide the Picture in picture icon in mobile view as it doesn't seem be supported in iOS, or android, and that could make more space for other icons.

ivanji:
Sure, thanks. BTW,  I'm going to push a couple more changes as I've been testing further and found some minor issues. I'll get rid of PiP in mobile view and bring back the sound control button, it's an accessibility issue not having it and there's no reason why it shouldn't be present.

pietrop:
ok, cool, thanks @ivanji !

re the sound toggle - ok that's good. 

A few other things I've noticed

**Progress bar**
In firefox it's a red dot while on chrome it's white?

**Current Timecode - colour**
Colour of current time, looks a bit faded `#E2A9A2` instead of `#a0372d` - was wondering what's the reason behind that?

**position of settings icon, and keyboard shortcuts icon**
Position of settings icon, and keyboard shortcuts icon in mobile view seems to go a bit off to the top of the page, rather then staying with the component, see below

Screen Shot 2019-03-14 at 22 37 06

ivanji:
I've fixed these issues and some minor ones I also noticed. I added Toggle Sound on/off button (instead of volume control as realistically that's what it is). 

> Colour of current time, looks a bit faded #E2A9A2 instead of #a0372d - was wondering what's the reason behind that?

Accessibility! previous red fails contrast checks. FYI, I didn't remove it, just added a new colour but like anything, it's to the devs what style they ultimately decide on, but hopefully always considering contrast. As a default option we might as well comply with best guidelines :)

Issue #113, opened 2019-03-14
Question and possible issue - initialising transcript component


ivanji:
Hi guys, so I've been experiencing an issue with local transcripts and hope you could help me understand a bit more what could be happening and if my approach is correct.

My previously edited local transcripts weren't loading, (due to the fact my mediaUrls contains a session signature appended to it which change periodically). After searching for localStorage load/save methods and updating this.props.mediaUrl to this.props.fileName, things still didn't work, unless I also update the componentDidUpdate lifecyle under TranscriptEditor.

Once I also updated ```prevProps.mediaUrl``` within this method my local transcripts started loading successfully. The problem then is that every time I started editing a transcript, the cursor started jumping to the beginning of the editor, and I could see in the console these repeated messages:

``` 
Transcript first and then media 
was already present in local storage...  
```

as the method keeps updating/executing it seems to be checking if the transcript is available locally - so I'm not understanding the purpose of why this code is under the componentDidUpdate method (or purpose of it, and the repetition under different conditions).

```js
componentDidUpdate(prevProps, prevState) {
    // Transcript and media passed to component at same time
    if (
      (prevState.transcriptData !== this.state.transcriptData)
        && (prevProps.mediaUrl !== this.props.fileName )
    ) {
      console.info('Transcript and media');
      this.ifPresentRetrieveTranscriptFromLocalStorage();
    }
    // Transcript first and then media passed to component
    else if (
      (prevState.transcriptData === this.state.transcriptData)
      && (prevProps.mediaUrl !== this.props.mediaUrl)
    ) {
      console.info('Transcript first and then media');
      this.ifPresentRetrieveTranscriptFromLocalStorage();
    }
    // Media first and then transcript passed to component
    else if (
      (prevState.transcriptData !== this.state.transcriptData)
      && (prevProps.mediaUrl === this.props.fileName)
    ) {
      console.info('Media first and then transcript');
      this.ifPresentRetrieveTranscriptFromLocalStorage();
    }
  }
``` 

don't know if these same changes (from mediaUrl -> fileName) caused an unexpected behaviour, but having this check running on every single update doesn't seem right.

I've already applied a solution to the issue by running the ```isPresentInLocalStorage``` method once during the componentDidMount lifecyle method instead, and I can edit my transcripts without hiccups. But I'm not sure if I'm missing something that you guys had in mind with the above code.

Don't know if this could be a section we could improve - and if there's scope also to provide an alternative prop to be used as title for localStorage as well.

[...]

PR #114, opened 2019-03-15
Fix 87 initialising transcript media v2

pietrop:

Recently noticed in PR https://github.com/bbc/react-transcript-editor/pull/111 but also present in current master and flagged in this  issue https://github.com/bbc/react-transcript-editor/issues/113

Originally logged as this issue https://github.com/bbc/react-transcript-editor/issues/87 and addressed as this PR https://github.com/bbc/react-transcript-editor/pull/88

**Describe what the PR does**    

When initialising the TranscriptEditor component you might have 

1. Both media url (audio or video) and transcript json present
2. Media url is present and transcript json is added later
3. Media url is added later and transcript json is present at initialisation

And it also needs to work, when saving to local storage and retrieving data from it.

In PR  https://github.com/bbc/react-transcript-editor/pull/88 fix all these scenarios but in current master `3` is not working.

This PR tries to figure out the differences between current master and previously fixed PR https://github.com/bbc/react-transcript-editor/pull/88

Not ready for merge, just using the diff to see what's the difference with current master. 

**Additional context**    

The idea is to rebase, and re-visit once these PRs have been added to master.

- PR https://github.com/bbc/react-transcript-editor/pull/111
- PR https://github.com/bbc/react-transcript-editor/pull/110


pietrop:
in `src/lib/TranscriptEditor/index.js` - function `ifPresentRetrieveTranscriptFromLocalStorage`

```diff
  ifPresentRetrieveTranscriptFromLocalStorage = () => {
-  if (this.refs.timedTextEditor!== undefined) {
-      if (this.refs.timedTextEditor.isPresentInLocalStorage(this.props.mediaUrl)) {
-        console.log('was already present in local storage');
-        this.refs.timedTextEditor.loadLocalSavedData(this.props.mediaUrl);
+    if (this.timedTextEditorRef.current!== undefined) {
+      if (this.timedTextEditorRef.current.isPresentInLocalStorage(this.props.mediaUrl)) {
+        console.info('was already present in local storage');
+        this.timedTextEditorRef.current.loadLocalSavedData(this.props.mediaUrl);
      } else {
        console.info('not present in local storage');
      }
    }
  }
```

in red how it was (when it was working), in green how it is current PR (not working)

pietrop:
Not ready for merge, just done further testing, doesn't work in conjunction with local storage.

See [QA docs `docs/qa/0-component-interface.md`](https://github.com/bbc/react-transcript-editor/blob/master/docs/qa/0-component-interface.md) for more info on how to test.

pietrop:
current issue, when initialising component in this branch:
- [x] video and transcript at same time --> works with local storage
- [ ] video first then transcript --> doesn't work with local storage 
- [ ] transcript first then video --> crashes 

crash error `TypeError: Cannot read property 'isPresentInLocalStorage' of null`

`src/lib/TranscriptEditor/index.js:83`

` if (this.timedTextEditorRef.current.isPresentInLocalStorage(this.props.mediaUrl)) {`

pietrop:
Digged out old PRs and Issues addressing this problem and fixed it.

Initial issue https://github.com/bbc/react-transcript-editor/issues/71 and PR https://github.com/bbc/react-transcript-editor/pull/72

`TranscriptEditor` component can be initialised either with  url `(1)`or from local media (blob)  `(2)` and it would restore session from local storage if previously saved.

When adding from local media, `TranscriptEditor` can take an optional `fileName` attribute (as explained in https://github.com/bbc/react-transcript-editor/pull/72 - https://github.com/bbc/react-transcript-editor/issues/71) which creates a unique key to save to local storage. 

See demo app in `src/index.js` line `36` for example of how to read file name in the browser client side to initialise the component with it.

```js
 // https://stackoverflow.com/questions/8885701/play-local-hard-drive-video-file-with-html5-video-tag
  handleChangeLoadMedia(files) {
    console.log(files);
    const file = files[0];
    const type = file.type;
    // check if is playable
    const videoNode = document.createElement('video');
    const canPlay = videoNode.canPlayType(type);
    if (canPlay) {
      const fileURL = URL.createObjectURL(file);
      // videoNode.src = fileURL
      this.setState({
        // transcriptData: kaldiTedTalkTranscript,
        mediaUrl: fileURL,
        fileName: file.name
      });
    }
    else {
      alert('select a valid audio or video file');
    }
  }
```

When testing `(1)` and `(2)`  note that even if we use it's "the same file", those are considered two different inputs by the component, as the component uses file name string for local media and the url uses the url string input to more uniquely identify transcriptions. 

Recognising whether is the same file, across url input string `(1)`, and local file `(2)`, it's outside of scope. Supporting local media is an edge case, we'd assume, and recommend most often the component be initialised with both media url and and transcript data (`json`) at the same time.

[QA doc -`0-component-interface.md`](https://github.com/bbc/react-transcript-editor/blob/master/docs/qa/0-component-interface.md) can be used forr more comprehensive test of all the different scenarios.

It might also be good to clean up this notes and add them the docs or make an ADR.

pietrop:
TL;DR: 

_to recap at high level_

When initialising the component there might be a race condition between media url + transcript Json content if not provided at same time

This can cause problem when retrieving from local storage

complicating things even further, if the media url comes from local media (and therefore as blob url ) then then url string is not a unique identifier that can be used to save and retrieve from local storage.

as a workaround added an optional file name to use local storage with local media

src/lib/TranscriptEditor/index.js
```diff
@@ -41,6 +49,8 @@ class TranscriptEditor extends React.Component {
 
   static getDerivedStateFromProps(nextProps) {
     if (nextProps.transcriptData !== null) {
+      console.log('getDerivedStateFromProps');
```
jamesdools: remove

src/lib/TranscriptEditor/index.js
```diff
@@ -52,43 +62,51 @@ class TranscriptEditor extends React.Component {
...
       this.ifPresentRetrieveTranscriptFromLocalStorage();
     }
   }
 
   ifPresentRetrieveTranscriptFromLocalStorage = () => {
-    if (this.timedTextEditorRef.current !== undefined) {
-      if (this.timedTextEditorRef.current.isPresentInLocalStorage(this.props.mediaUrl)) {
+    console.log('checking - ifPresentRetrieveTranscriptFromLocalStorage');
+    if (
+      this.timedTextEditorRef !== undefined &&
```
jamesdools: I think this can just be written as

```
const editor = this.timedTextEditorRef;

if (editor && editor.current) {
  if (editor.current.isPresentInLocalStorage(this.props.mediaUrl)) {
     console.info('Already present in local storage.');
     this.timedTextEditorRef.current.loadLocalSavedData(this.props.mediaUrl);	        
  } else {
    console.info('Not present in local storage.');
  }

}

```

src/lib/TranscriptEditor/index.js
```diff
@@ -52,43 +62,51 @@ class TranscriptEditor extends React.Component {
...
     ) {
       console.info('Media first and then transcript');
       this.ifPresentRetrieveTranscriptFromLocalStorage();
     }
   }
 
   ifPresentRetrieveTranscriptFromLocalStorage = () => {
-    if (this.timedTextEditorRef.current !== undefined) {
-      if (this.timedTextEditorRef.current.isPresentInLocalStorage(this.props.mediaUrl)) {
+    console.log('checking - ifPresentRetrieveTranscriptFromLocalStorage');
```
jamesdools: remove

src/lib/TranscriptEditor/index.js
```diff
@@ -52,43 +62,51 @@ class TranscriptEditor extends React.Component {
...
-      && (prevProps.mediaUrl === this.props.mediaUrl)
+      prevState.transcriptData !== this.state.transcriptData &&
+      prevProps.mediaUrl === this.props.mediaUrl
     ) {
       console.info('Media first and then transcript');
       this.ifPresentRetrieveTranscriptFromLocalStorage();
     }
   }
 
   ifPresentRetrieveTranscriptFromLocalStorage = () => {
```
jamesdools: is there a better name for this..?

src/lib/TranscriptEditor/index.js
```diff
@@ -99,25 +117,25 @@ class TranscriptEditor extends React.Component {
...
     const currentTime = e.target.currentTime;
     this.setState({
-      currentTime,
+      currentTime
     });
-  }
+  };
 
-  handlePlayMedia = (bool) => {
+  handlePlayMedia = bool => {
```
jamesdools: I must have missed this from before - but can we not do this. bool is the worst name 🙃

jamesdools: We've changed this elsewhere I believe. so keep things in line with the rest of the codebase.

https://github.com/bbc/react-transcript-editor/blob/ad9526fba90a3fec5cfaef12b4540238f8f9abbf/src/lib/TranscriptEditor/MediaPlayer/PlayerControls.js#L147

we use `isPlaying` here, and propTypes declares it as a boolean. `bool` isn't useful because the true or false value doesn't describe what state it should take.

jamesdools: Sidenote - I think this is bad logic that elsewhere, it's a function: https://github.com/bbc/react-transcript-editor/blob/ad9526fba90a3fec5cfaef12b4540238f8f9abbf/src/lib/TranscriptEditor/TimedTextEditor/index.js#L548

one way or the other. needs some refactoring

src/lib/TranscriptEditor/index.js
```diff
@@ -157,18 +174,16 @@ class TranscriptEditor extends React.Component {
...
 
-    this.setState({ timecodeOffset: timecodeOffset },
-      () => {
-        // eslint-disable-next-line react/no-string-refs
-        this.timedTextEditorRef.current.forceUpdate();
-      });
-  }
+  handleSetTimecodeOffset = timecodeOffset => {
+    this.setState({ timecodeOffset: timecodeOffset }, () => {
+      // eslint-disable-next-line react/no-string-refs
```
jamesdools: this should be able to go now

src/lib/TranscriptEditor/index.js
```diff
@@ -250,149 +268,194 @@ class TranscriptEditor extends React.Component {
...
+        {this.props.mediaUrl === null ? null : header}
 
         <div className={ style.grid }>
           <section className={ style.row }>
             <aside className={ style.aside }>
               {this.props.mediaUrl === null ? null : videoPlayer}
             </aside>
             <main className={ style.main }>
-              {this.props.mediaUrl === null ? null : timedTextEditor}
+              {this.props.mediaUrl !== null &&
```
jamesdools: not a fan of this but ok

src/lib/TranscriptEditor/index.js
pietrop: did you try it? does it work?

pietrop: ok trying to change to 

```js
 ifPresentRetrieveTranscriptFromLocalStorage = () => {
    const timedTextEditor = this.timedTextEditorRef;
    if (timedTextEditor && timedTextEditor.current) {
      if (timedTextEditor.current.isPresentInLocalStorage(this.props.mediaUrl)) {
        console.info('Already present in local storage.');
        timedTextEditor.current.loadLocalSavedData(this.props.mediaUrl);
      } else {
        console.info('Not present in local storage.');
      }
    }
  };
```

I prefer `timedTextEditor` coz is less ambiguous compared to `editor`.

pietrop: ok, tested it works

src/lib/TranscriptEditor/index.js
pietrop: for `ifPresentRetrieveTranscriptFromLocalStorage`?

src/lib/TranscriptEditor/index.js
pietrop: ok, changed it to `isPlaying`, well spotted.

src/lib/TranscriptEditor/index.js
pietrop: ok we can refactor later

Issue #109, opened 2019-03-11
Restructuring the repo for easier componentisation


chrishutchinson:
Hi all. I had a great chat with @pietrop today about this project, as I'm keen to contribute and use it at The Times. A few things came up, particularly around making the sub-components of the app available for re-use (as I'll want to bring a lot of my own UI and other logic for our use case, but would like to share the common bits). This issue is to discuss an approach to doing this.

**Is your feature request related to a problem? Please describe.**
As above, I'd like to make the sub-components of this app available for re-use. At present, it's pretty hard to only take the text editor, or the media player, without bringing the entire rest of the application.

**Describe the solution you'd like**
To achieve this, I'm proposing we transition the repo to a monorepo (powered by [Lerna](https://github.com/lerna/lerna) and [Yarn Workspaces](https://yarnpkg.com/lang/en/docs/workspaces/)) with a small number of sub-packages, each of which can be pushed to NPM (or equivalent) as independent React components. This would allow for a consuming app (like mine) to install the components it needs, and bring it's own for the areas where either:
a) it needs something bespoke (e.g. user authentication, custom settings UI)
a) bbc/react-transcript-editor doesn't provide that feature, and it doesn't make sense for the overall roadmap
c) I want to replace a component for a one-off use case

Having tinkered with this, I roughly imagine we'd have the following packages in the monorepo:
- `@bbc-transcript-editor/app`: A fully composed version of the app, which can be rendered in a standard `create-react-app` and allow someone with little-to-no desire the customise the app to get started
- `@bbc-transcript-editor/media-player`: The media player, including all play/pause controls etc.
- `@bbc-transcript-editor/settings`: The settings modal
- `@bbc-transcript-editor/timed-text-editor`: The core DraftJS-based transcript editor
- `@bbc-transcript-editor/adapter-{FROM}-{TO}`: A standardised set of adapters, replacing the existing adapter / converter logic (e.g. BBC Kaldi to DraftJS, or DraftJS to SRT).
- `@bbc-transcript-editor/context`: A `React.Context` wrapper, the root of all instances of the Editor, which provides default global config options, allows developers to register active adapters, and handles the globally-utilised analytics callbacks. (To do this, all components would need to be updated so they were context-aware. This would be done more easily if we upgraded to React 16.8, as we would have access to the `useContext` hook).
[...]

PR #121, opened 2019-03-19
refactor demo navbar

pietrop:

Temporary fix before more substantial refactor https://github.com/bbc/react-transcript-editor/issues/109

**Describe what the PR does**    

@jamesdools as discussed, here's a refactor of the demo page navbar.

Ready for review

**Additional context**    
Desktop and mobile

Screen Shot 2019-03-19 at 11 20 53

Screen Shot 2019-03-19 at 11 21 13

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -149,58 +149,90 @@ class App extends React.Component {
            >
             Github Repo
            </a>
+           <hr/>
          </span>
-         <br />
-         <button onClick={ () => this.loadDemo() }>load demo</button>
-         <hr />
-         <label>Load Local Media</label>
-         <input
-           type="file"
-           onChange={ e => this.handleChangeLoadMedia(e.target.files) }
-         />
+
+         <div className={ style.demoNav }>
+           <section className={ style.demoNavItem }>
+             <button onClick={ () => this.loadDemo() }>load demo</button>
+
+           </section>
+           <section className={ style.demoNavItem }>
+
+             <label>Load Local Media</label>
+             <br/>
+             <input
+               className={ style.demoInput }
+               type="file"
+               onChange={ e => this.handleChangeLoadMedia(e.target.files) }
+             />
+             <br/>
          or
-         <button onClick={ () => this.handleChangeLoadMediaUrl() }>
+             <br/>
+             <button onClick={ () => this.handleChangeLoadMediaUrl() }>
           Load Media From Url
-         </button>
-         <br/>
-         <label>open Transcript Json</label>
-         <SttTypeSelect
-           name={ 'sttType' }
-           value={ this.state.sttType }
-           handleChange={ this.handleSttTypeChange }
-         />
-         <input
-           type="file"
-           onChange={ e => this.handleChangeLoadTranscriptJson(e.target.files) }
-         />
+             </button>
 
-         <br />
-         <label>Export transcript</label>
-         <button onClick={ () => this.exportTranscript() }>Export file</button>
+           </section>
+           <section className={ style.demoNavItem }>
 
-         <ExportFormatSelect
-           name={ 'exportFormat' }
-           value={ this.state.exportFormat }
-           handleChange={ this.handleExportFormatChange }
-         />
-         <br />
-         <label>Text Is Editable</label>
-         <label>
-           <input type="checkbox"
-             defaultChecked="true"
-             onChange={ this.handleIsTextEditable }
-           />
-         </label>
-         <br/>
-         <label>Optional Transcript Name</label>
-         <input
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -71,158 +71,188 @@ class App extends React.Component {
     if (file.type === 'application/json') {
       const fr = new FileReader();
 
-      fr.onload = (evt) => {
+      fr.onload = evt => {
```
jamesdools: why evt? consistency with the rest of the file please 🙏

src/index.js
```diff
@@ -71,158 +71,188 @@ class App extends React.Component {
...
-  }
+    this.setState(prevState => ({
+      isTextEditable: prevState.isTextEditable !== true
+    }));
+  };
 
   // https://stackoverflow.com/questions/21733847/react-jsx-selecting-selected-on-selected-select-option
-  handleSttTypeChange = (event) => {
+  handleSttTypeChange = event => {
     console.log(event.target.name, event.target.value);
```
jamesdools: remove these logs

src/index.js
```diff
@@ -71,158 +71,188 @@ class App extends React.Component {
...
-       </div>
-     );
-   }
+          </a>
+          <hr />
+        </span>
+
+        <div className={ style.demoNav }>
+          <section className={ style.demoNavItem }>
+            <button onClick={ () => this.loadDemo() }>load demo</button>
```
jamesdools: Some consistency with text formatting would make this all look cleaner.

`load demo`
`Load Local Media`
`open Transcript`
`Export transcript`

the capitalisation is all over the place and makes it look a lil messy

src/index.js
```diff
@@ -71,158 +71,188 @@ class App extends React.Component {
...
+            <br />
+            <ExportFormatSelect
+              name={ 'exportFormat' }
+              value={ this.state.exportFormat }
+              handleChange={ this.handleExportFormatChange }
+            />
+            <br />
+            <button onClick={ () => this.exportTranscript() }>Export file</button>
+          </section>
+          <section className={ style.demoNavItem }>
```
jamesdools: do we still need text is editable?!

src/index.js
```diff
@@ -8,7 +8,8 @@ import style from './index.module.css';
 import SttTypeSelect from './select-stt-json-type';
 import ExportFormatSelect from './select-export-format';
 
-const tedTalkVideoUrl = 'https://download.ted.com/talks/KateDarling_2018S-950k.mp4';
+const tedTalkVideoUrl =
```
jamesdools: Did the linter do this? it's not crazy long so I'd prefer it on one line

src/index.js
```diff
@@ -71,158 +71,188 @@ class App extends React.Component {
...
+  };
+
+  handleChangeTranscriptName = value => {
+    this.setState({ fileName: value });
+  };
+
+  render() {
+    return (
+      <div className={ style.container }>
+        <span>
```
jamesdools: This span doesn't look great. Can make the whole thing a link instead of GitHub Repo as a separate url -

or alternatively a github icon linking to it. Also maybe center text.

I kinda think it's not necessary at all in it's current form so happy to chuck it 🤷‍♂️

src/index.js
pietrop: yeah, must have been the linter, or prettier..

pietrop:
Yeah, I think `package-lock.json` was removed in previous commits. But since npm 5 it's good practice to include it as npm will handles any possible conflict itself.

pietrop:
I'd still keep Text Is Editable, eg think situations like the BBC Volt Use case, where v1 is not editable and v2 is. Or there might be a situation where you have an admin user that edits the text and another user that can only view it but not edit etc..

src/index.js
pietrop: good catch, it was from a copy and paste 🤦‍♂️

src/index.js
pietrop: yeah, I'd still keep Text Is Editable, eg think situations like the BBC Volt Use case, where v1 is not editable and v2 is. Or there might be a situation where you have an admin user that edits the text and another user that can only view it but not edit etc.. Or we do another prototype, different use case and you want to have the "interactive transcript" but non editable (eg is a new story format) etc..

src/index.js
pietrop: I've replaced it with the github icon, and removed the span. But we need something saying that that page is a demo otherwise some users might get confused etc..
+ good to link to the repo in case someone finds the project through the demo page etc..

pietrop:
ok, done the changes, ready for merge @jamesdools

PR #122, opened 2019-03-19
Feature: Introduce Storybook + monorepo structure

chrishutchinson:
**Is your Pull Request request related to another [issue](https://github.com/bbc/react-transcript-editor/issues) in this repository ?**      
Initially raised in issue #109.

**Describe what the PR does**    
This PR overhauls the structure of the application, otherwise keeping everything else intact. The reasons for this change are to support more modular use of the components for users who wish to bring their own features and UI, while taking advantage of the great work inside this codebase.

A fairly comprehensive list of changes is below:
- [x] Move most components, utilites, adapters and exporters into logical top-level packages (some contain sub-components, where there is no need for them to be shared)
- [x] Introduce Lerna for managing the monorepo, alongside Yarn Workspaces which is required for it to run (`npm` may not work..)
- [x] Introduce Storybook, used primarily for exposing the new React packages in a local development environment (although it also creates a nice static site for demoing!)
- [x] Run `prettier` on _all_ JavaScript files, and simplifies the `eslint` config so there are not conflicts with `prettier`s defaults
- [x] Scopes all packages under the `@bbc-transcript-editor/*` namespace, which would need to become a new organisation in NPM

**State whether the PR is ready for review or whether it needs extra work**    
This PR is absolutely ready for review, but also very possibly will need further work. That entirely depends 😃 

**Additional context**
I believe there will be some additional work in this PR, specifically:

- [ ] Agreement, then BBC approval, of transitioning to a new NPM organisation (`@bbc-transcript-editor/*`) for the packages
- [ ] Extensive testing, to make sure I haven't broken any specific logic while moving things around
- [ ] Finalise changes to the `README`
- [ ] Agreement on the terms for `adapters` and `exporters` (still not sure on this)
- [ ] Updating the `travis` config to support automated packaging and deployment to NPM on merge to `master`
- [ ] Determine how to bundle + ship each package to `NPM`, as an ES6 module, or as a transpiled ES5 component

One this PR has been merged, there may be some additional work, notably:

- [ ] Introducing a new "wrapper" component to centrally manage analytics events, and registering which adapters + exporters to use (useful for those that want to create custom interfaces using some of these components)

Screen Shot 2019-03-19 at 20 31 37
Screen Shot 2019-03-19 at 20 31 46
Screen Shot 2019-03-19 at 20 31 48
Screen Shot 2019-03-19 at 20 31 52
Screen Shot 2019-03-19 at 20 31 54
Screen Shot 2019-03-19 at 20 31 57

PR #123, opened 2019-03-20
Stt adapter IBM - andrew d anderson

pietrop:

from original PR https://github.com/bbc/react-transcript-editor/pull/60 by @AndrewDAnderson

**Describe what the PR does**    

Adds support for IBM Watson STT (cc @seanmtracey)

Updated PR https://github.com/bbc/react-transcript-editor/pull/60 with current master, and doing some final tweaks before merging to master, ahead off reorg (PR https://github.com/bbc/react-transcript-editor/pull/122 - issue https://github.com/bbc/react-transcript-editor/issues/109) 

**Additional context**    
NA


pietrop:
Need to double check the tests are passing, but other then that ready for merge to master.

cleaned up the code making it more in line with the logic in the other adapter, and better setup for accommodating changes and adjustment over time.

Seems like IBM speaker diarization might actually be quiet good, in this example it recognised it as being just one speaker. 

So rather then grouping by speaker, which would have left with one giant paragraph and indured draftjs performance, with virtual dom etc.. I left some comments in the code on where that could be changed, eg it could break paragraphs every minute.

This also loses some of the clever stuff that was done in the previous PR about adding punctuation to IBM STT which doesn't seem to have it (?).

[Seems like a common workaround is to add punctuation after every IBM line](https://developer.ibm.com/answers/questions/322036/watson-speech-to-text-punctuation-1/) but will leave that for a later PR.

pietrop:
fixed the test, ready for merge 

Screen Shot 2019-03-20 at 19 49 20

src/lib/Util/adapters/ibm/index.js
```diff
@@ -0,0 +1,232 @@
...
+  };
+
+  const ibmNormalisedWordsToDraftJs = (ibmNormalisedWordsWithSpeakers) => {
+    const drafJsParagraphsResults = [];
+    ibmNormalisedWordsWithSpeakers.forEach((ibmParagraph) => {
+      const draftJsContentBlockParagraph = {
+        text: ibmParagraph.map((word) => {return word.text;}).join(' '),
+        type: 'paragraph',
+        data: {
+          // TODO: Assuming each paragraph in IBM line is the same
```
jamesdools: does more work need to be done here?

src/lib/Util/adapters/ibm/index.js
```diff
@@ -0,0 +1,232 @@
+/**
```
jamesdools: Let's remove these comments - we have some sample json files to refer to if needed.

src/lib/Util/adapters/ibm/index.js
```diff
@@ -0,0 +1,232 @@
...
+
+        word.speaker = findSpeakerSegmentForWord(word, ibmSpeakers);
+
+        return word;
+      });
+    });
+  };
+
+  const ibmNormalisedWordsToDraftJs = (ibmNormalisedWordsWithSpeakers) => {
+    const drafJsParagraphsResults = [];
```
jamesdools: let's just correct this variable name.

src/lib/Util/adapters/ibm/index.js
```diff
@@ -0,0 +1,232 @@
...
+      };
+      drafJsParagraphsResults.push(draftJsContentBlockParagraph);
+    });
+
+    return drafJsParagraphsResults;
+  };
+
+  const normalisedWords = normalizeIBMWordsList(ibmJson.results[0].results);
+  // nested array of words, to keep some sort of paragraphs, in case there's only one speaker
+  // can be refactored/optimised later
```
jamesdools: Will it? Let's make this comment a `TODO:`'s and actually follow up on it!

Slightly side note but there's lots of comments like this in the code and it gets lost. Even though there are a fair few TODOs too 😅

src/lib/Util/adapters/ibm/index.js
pietrop: Don't think so, It just seems like the IBM data structure gives you word level speakers,but also gives you "lines" so assuming each word in a line has the same speaker.

Can also check with @seanmtracey if has time to review this later.

pietrop: removing the todo and expanding on the comment to give more context.

src/lib/Util/adapters/ibm/index.js
pietrop: well spotted

src/lib/Util/adapters/ibm/index.js
pietrop: ok ,fair added the TODO

src/lib/Util/adapters/ibm/index.js
pietrop: ok, removed across all the other adapters as well

PR #128, opened 2019-04-05
move TimedTextEditor, TranscriptEditor and adapters util to packages

pietrop:

A spike on branch [monorepo-refactor](https://github.com/bbc/react-transcript-editor/tree/monorepo-refactor)

**Describe what the PR does**    
- move TimedTextEditor and TranscriptEditor to packages  
- as well as adapters dependencies from `Util`

still working on it

**Additional context**    
🚀


pietrop:
Got the storybook working! based on @chrishutchinson  PR  https://github.com/bbc/react-transcript-editor/pull/122


There's still stuff to do, but as short lived spike, I think it be good to review it and merge it to the mono repo branch, before doing another spike with some more tweaks. 

so far done
- [x] Moved everything in src/lib into `packages`
- [x] config lerna so that it can pick up nested folder structure inside of packages
- [x] demo page - The advantage of keeping. refactoring demo page, is that you can also have a direct link to full screen version, eg publishing in github pages [http://localhost:6006/iframe.html?id=demo--default](http://localhost:6006/iframe.html?id=demo--default)


Still to do, for another spike, PR

- [ ] [folder structure for packages folder - see this example](https://github.com/bbc/react-transcript-editor/pull/122#issuecomment-475216748)
- [ ] re-think the naming of the modules, eg is this a good naming convention `@bbc/react-transcript-editor-timecode-converter` as a workaround to the name spacing? or is it not very good? do we even need to name space them? we could just decide which once are public eg they are reusable by themselves, so they are published to npm , and in which case what is a sensible standalone name, and which once are private, meaning used only within the component) - this could be in more then one pass.
- [ ] figure out how to publish to npm only certain packages with lerna.
- [ ] figure out how to publish storybook to github pages

pietrop:
some quick lerna tips, that we should add to docs notes

use `npx` to call local version of lerna, as it's installed as dev dependency 

```
npx lerna ls
```

lists all the modules it recognises in `packages` folder and other folders specified in `lerna.json` config. useful to see if it is picking up everything you expect.

```
npx lerna bootstrap
```
links all the `packages` together as needed.

if that works, you should then be ready to do 

```
npm run storybook 
```

should take you to [http://localhost:6006](http://localhost:6006) where you should see the storybook. unless any errors are flagged in the terminal.

PR #130, opened 2019-04-05
Monorepo refactor

pietrop:

https://github.com/bbc/react-transcript-editor/pull/122

 **Describe what the PR does**    
Restructures the codebase to be more modular, introducing storybook

still in draft


pietrop:
I am sold on the storybook, but some things to figure out are, do we really need Lerna?

Argument for removing Lerna in the monorepo setup. 

1.  _npm7 is likely to have workspaces_

Not only npm latest version is head to head with yarn features, but also [npm@7, which is predicted for late summer and fall of 2018 is very likely to have Workspaces](https://blog.npmjs.org/post/173239798780/beyond-npm6-the-future-of-the-npm-cli)

imho lerna is poorly documented, and hard to get up and running with. 

using lerna would also require rest of our team to wrap their heads around that for this specific project, has its not currently part of our stack, and might limit the number of contributions etc.

most importantly tho, solves a problem that we don't have for this specific project. Lerna is for managing monorpeos where you publish to npm separate packages to be installed as standalone. 

Eg In our repo we are leaning torwards publishing one package @bbc/react-transcript-editor and letting you import the parts that are needed for your project.

Which leads me to my second point below.

2. _Can export various components from the same npm module_

With [React Bootstrap for example](https://react-bootstrap.github.io/getting-started/introduction) 

>You should import individual components like: react-bootstrap/Button rather than the entire library. Doing so pulls in only the specific components that you use, which can significantly reduce the amount of code you end up sending to the client.

In other words when you install `react-bootstrap` with npm, if for example you only want a button.  doing  `import Button from 'react-bootstrap/Button';` allows you to only bundle the btn and not the rest of the `react-bootstrap` library.

Will do a separate spike form this branch/PR to show what it would look like without lerna.

Issue #127, opened 2019-04-05
Support for AWS Transcribe Speaker Detection


philmcmahon:
**Is your feature request related to a problem? Please describe.**
Currently the AWS Transcribe adaptor doesn't pick up the AWS speaker detection stuff - each line gets listed as a separate speaker. We're looking into doing trasnscription stuff at the guardian at the moment and I heard about this tool from Rich Beddington in the editorial tools team here. Speaker detection for AWS would be a nice feature so I was thinking of having a go at adding it. 

**Describe the solution you'd like**
Text grouped by speaker in the transcript, rather than each line listed as TBC 0. The data for this appears to already be available in the json output by AWS so it should be achievable.

**Describe alternatives you've considered**
Nothing as yet.

**Additional context**
I've got some time next week to work on this, any suggestions appreciated - do you think this is worth doing?

pietrop:
👋  @philmcmahon, Thanks for reaching out, I had also left a message on your https://github.com/guardian/transcribe-tool/issues/1 😄 

I think this is worth doing, and yes, you can see in this PR https://github.com/bbc/react-transcript-editor/pull/120 that the speaker segmentation was left for a later stage.

At the moment me and @jamesdools are addressing this issue https://github.com/bbc/react-transcript-editor/issues/109 following @chrishutchinson PR example https://github.com/bbc/react-transcript-editor/pull/122 to make the codebase more modular, using a monorepo with storybook.

There's a [Guide: How to Create an Adapter](https://github.com/bbc/react-transcript-editor/blob/master/docs/guides/adapters.md) you can have a look at, but any questions feel free to reach out.

In the current master setup you can find the adapters in `src/lib/Util/adapters/` as explained in the guide above. 

In the monorepo/storybook refactor, it will end up somewhere inside the `packages` folder. 

But because the adapters are pretty self contained, you could get started in the current setup and then move it later if we are done with the storybook PR before then, if that makes sense. And we can always help out if it gets confusing.

philmcmahon:
Thanks @pietrop - I've left it with the current setup for now as it's just a couple of files with some fixtures so should be fairly easy to move.

pietrop:
Thanks @philmcmahon I've also added it to the repo refactor PR https://github.com/bbc/react-transcript-editor/pull/135 so that it shouldn't be effected by that, when that's ready for merge

PR #131, opened 2019-04-07
Add amazontranscribe speaker detection support

philmcmahon:
Resolves https://github.com/bbc/react-transcript-editor/issues/127 

**Describe what the PR does**    
This adds support for speaker detection for the `amazontranscribe` adaptor. It's fairly similar to how this works for the BBCKaldi adaptor. One thing I wasn't sure of was how to enable/disable the speaker grouping, so I've currently set it based off whether there's a speaker_labels block in the amazon JSON. 

Ready for review, very happy to make changes though!

**Additional context**    
Currently this doesn't look fantastic in the UI as the speaker label gets cropped, but that's perhaps for a future PR.
Screenshot from 2019-04-07 20-57-20


pietrop:
Looks good, thanks @philmcmahon  for writing the tests, and handling both when speaker segments are present as well as when they are not.

PR #140, opened 2019-04-09
Don't add 'speaker' prefix twice.

philmcmahon:
**Describe what the PR does**    
Fixes bug where the speaker labels end up being something like 'Speaker Speaker 0' as opposed to just 'Speaker 0', as a result of the 'Speaker ' prefix being added both in `index.js` and in `group-words-by-speakers.js`

Ready for review

**Additional context**    
Oops :(


pietrop:
Looks good, thanks @philmcmahon

PR #155, opened 2019-05-08
add spellCheck support

terite:
No

**Describe what the PR does**    
This PR adds support for passing the `spellCheck` prop to the draft-js Editor component, and enhances the demo to allow passing that prop.

**Additional context**    
* https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/spellcheck
* https://draftjs.org/docs/api-reference-editor#spellcheck

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -20,6 +20,7 @@ class App extends React.Component {
       transcriptData: null,
       mediaUrl: null,
       isTextEditable: true,
+      spellCheck: false,
       sttType: 'bbckaldi',
       analyticsEvents: [],
       title: '',
@@ -83,10 +84,16 @@ class App extends React.Component {
     }
   }
 
-  handleIsTextEditable = () => {
-    this.setState(prevState => ({
-      isTextEditable: prevState.isTextEditable !== true
-    }));
+  handleIsTextEditable = (e) => {
+    this.setState({
+      isTextEditable: e.target.checked
+    });
+  };
+
+  handleSpellCheck = (e) => {
+    this.setState({
+      spellCheck: e.target.checked
+    });
   };
 
   // https://stackoverflow.com/questions/21733847/react-jsx-selecting-selected-on-selected-select-option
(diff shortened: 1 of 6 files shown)


demo/app.js
```diff
@@ -226,11 +233,21 @@ class App extends React.Component {
               <input
                 id={ 'textIsEditableCheckbox' }
                 type="checkbox"
-                defaultChecked="true"
+                checked={ this.state.isTextEditable }
```
terite: A controlled component seems like the way to go here.

https://reactjs.org/docs/forms.html#handling-multiple-inputs
https://reactjs.org/docs/uncontrolled-components.html#default-values

packages/components/timed-text-editor/index.js
```diff
@@ -509,6 +509,7 @@ class TimedTextEditor extends React.Component {
           blockRendererFn={ this.renderBlockWithTimecodes }
           handleKeyCommand={ command => this.handleKeyCommand(command) }
           keyBindingFn={ e => this.customKeyBindingFn(e) }
+          spellCheck={ this.props.spellCheck }
```
pietrop:
```suggestion
          spellCheck={ this.props.spellCheck?  this.props.spellCheck : false  }
```

What do you think about having a default to false if the prop is not specified to making it an optional parameter to the component?

pietrop:
Just noticed it works in Chrome, doesn't seem to work in Firefox? (on a mac)

terite:
I'm able to get spell checking on macOS 10.14.4 using Firefox `66.0.5 (64-bit)`


Does FireFox have a dictionary pre-installed for your locale? https://support.mozilla.org/en-US/kb/how-do-i-use-firefox-spell-checker

In any case, actual spell checking is browser dependent and out of my control as per links in the description.

packages/components/timed-text-editor/index.js
terite: This parameter is already optional. `this.props.spellCheck` will evaluate to `true`, `false`, or `undefined`. All 3 are valid values. Additionally, the relevant prop type is not suffixed with `isRequired`.

Issue #159, opened 2019-05-15
Unecessary re-renders


pietrop:
**Describe the bug**
Using `Highlighting Component Updates` in React Chrome Dev tool, gives some insights on parts of the component that might be re-rendering unnecessarily.

Possibly linked to issue https://github.com/bbc/react-transcript-editor/issues/150

**To Reproduce**
1. Go to 'Demo app'
2. Click on 'Load Demo'
3. open React dev tools, click on settings, and check box for 'highlight updates'
3. click 'play' btn 
4. See colours hilights  

**Expected behavior**

Fewer re-renders, and no re-render for components not involved in the playback (eg ToolTip)

**Screenshots**

![performance-issue](https://user-images.githubusercontent.com/4661975/57791633-158c5080-7735-11e9-8522-150bbd2946f0.gif)

**Desktop (please complete the following information):**
 - OS:`OSX 10.13.6`
 - Browser `Chrome 74.0.3729.131`
 - Version `1.0.2`

**Additional context**

- [Make React Fast Again [Part 3]: Highlighting Component Updates](https://blog.logrocket.com/make-react-fast-again-part-3-highlighting-component-updates-6119e45e6833)

pietrop:
Explored in branch PR/Branch https://github.com/bbc/react-transcript-editor/pull/160

![performance-issue-solution](https://user-images.githubusercontent.com/4661975/57796498-46be4e00-7740-11e9-8685-37ea2f503817.gif)

Seem to have found a way to reduce unnecessary re-renders using should component update 🚀

pietrop:
addressed in https://github.com/bbc/react-transcript-editor/pull/160  and currently in https://github.com/bbc/react-transcript-editor/pull/175 soon to be merged into master, so closing for now

Issue #150, opened 2019-04-26
Performance hit for media over 1 hour 


pietrop:
# **Describe the bug**

There seems to be a noticeable performance hit for media over 1 hour - in this case 2h 42 min video.

# **To Reproduce**
## Prep - get a transcript for along file
_these are the steps I followed, irrelevant for the bug itself, but helpful to setup problem_
- I got this video [from vimeo](https://vimeo.com/36579366)
- used [pietrop/electron-video-downloader](https://github.com/pietrop/electron-video-downloader) to get the video, (electron wrapper to youtubedl)
- created a video over 2 hours, closer to 3 by concating shorter video with ffmpeg ([ffmpeg-remix](https://www.npmjs.com/package/ffmpeg-remix) on an mp4)
- create transcript using [autoedit.io](http://autoedit.io) using pocketsphinx option for STT
- exported autoEdit2 json once transcript is ready

## Steps to reproduce the behavior:
1. get a long media and STT json transcript
1. Go to [storybook demo page](https://bbc.github.io/react-transcript-editor/?path=/story/demo--default)
2. Click on 'import transcript' select 'autoEdit2'
_This is the json I used
[three-hours-sample.mp4_26_04_2019_15-12-51.json.zip](https://github.com/bbc/react-transcript-editor/files/3121678/three-hours-sample.mp4_26_04_2019_15-12-51.json.zip)_
2. Click on 'play'
3. Scroll down to text 
2. Click on a word, and start typing
4. See error - there's a noticeable lag between when you type and when you see it on screen

# **Expected behaviour**

No lag for videos over 1 hour, same responsiveness as for video under 1 hour.

# **Screenshots**

![performance-hit-over-1-hour](https://user-images.githubusercontent.com/4661975/56814837-e66e7780-6837-11e9-9542-08043575f949.gif)

# **Desktop (please complete the following information):**
 - OS: Mac OSX 10.13.6 - High Sierra
 - Browser: Chrome Version 73.0.3683.103 (Official Build) (64-bit)
 - Version: `@bbc/react-transcript-editor@0.5.1`

# **Additional context**

Not sure what's the issue 🤷‍♂️ 

Might be mostly to do with draftjs?

https://github.com/facebook/draft-js/issues/437

Laurian:
[...]

Issue #147, opened 2019-04-15
`TimedTextEditor` Text centre aligned in mobile view


pietrop:
**Describe the bug**

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to [Storybook demo app](https://bbc.github.io/react-transcript-editor/?path=/story/demo--default)
2. Click on `load demo`
2. Click on `screen size icon` and chose `iphone 6`
3. Scroll down to see `TimedTextEditor`
4. See error, text is centred align

**Expected behavior**

Expect text to be aligned to the left.

**Screenshots**

![Screen Shot 2019-04-15 at 16 39 40](https://user-images.githubusercontent.com/4661975/56145967-485bf100-5f9d-11e9-9b9b-5699b49fdd7e.png)

**Desktop (please complete the following information):**
 - OS: Mac, OS X,  Storybook - iPhone6 view
 - Browser : Chrome

**Additional context**

Seems to be caused by `/demo/index.module.css` under

```css
/* phone */
@media (max-width: 767px) {
  body {
    padding: 0;
    text-align: center;
  }
...
```

Which is effecting the all of the page, if commenting out ` text-align: center;` then error goes away.

![Screen Shot 2019-04-15 at 16 50 00](https://user-images.githubusercontent.com/4661975/56146678-89a0d080-5f9e-11e9-8b40-a6b889c0a61a.png)

pietrop:
closing as fixed in https://github.com/bbc/react-transcript-editor/pull/175 PR soon to be merged into master.

PR #160, opened 2019-05-15
remove unnecessary re-renders to improve performance

pietrop:

This issue https://github.com/bbc/react-transcript-editor/issues/159 

**Describe what the PR does**    
Moved `header` into a component with `shouldComponentUopdate` false to test avoid unecessary re-renders - test

_work in progress_

**Additional context**


pietrop:
Trying out [`why-did-you-update`](https://github.com/maicki/why-did-you-update)

Screen Shot 2019-05-17 at 17 34 59

Screen Shot 2019-05-17 at 17 33 12


Screen Shot 2019-05-17 at 17 33 55

Screen Shot 2019-05-17 at 17 34 05

See [Common Fixing Scenarios](https://github.com/maicki/why-did-you-update#common-fixing-scenarios)


- [ ] [Changes Are in Functions Only](https://github.com/maicki/why-did-you-update#changes-are-in-functions-only)
- [ ] [Value Did Not Change](https://github.com/maicki/why-did-you-update#value-did-not-change)

pietrop:
[Rough draft of React performances optimisation notes from various docs and blogs can be found here](https://github.com/bbc/react-transcript-editor/blob/fix-159-performance-problem/docs/notes/2019-05-16-prevent-unnecessary-re-renders-in-react.md)

pietrop:
To review this PR you need to run npm install as I’ve added `react-visibility-sensor` in `package.json`.

## `currentTime` in `TimedTextEditor`

 Updating on currentTime in TimedTextEditor might have a performance cost?

Or at the very lease it causing frequent re-renders

 ```js  if (nextProps.currentTime !== this.props.currentTime ) {
      return true;
    } ```  

18ms + 17 to 24 re-renders calls

Vs 
0.3ms + 17 to 24 re-renders calls if not updating TimedTextEditor on currentTime change (eg commenting out those 3 lines)

In current setup, draftJS Editor re-render should be separated from css injection in TimedTextEditor to use currentTime to show current word to at least avoid re-render DraftEditor on `currentTime` change.

## `react-visibility-sensor`

Following  up [from this comment](https://github.com/bbc/react-transcript-editor/issues/150#issuecomment-487176355) from @laurian added `VisibilitySensor` in WrapperBlock, around the Speaker, Timecode, and draftJs `Editor`. This should avoid rendering any draftjs block unless it is visible

Downside of using `VisibilitySensor` is the text not in viewport is not rendered. Which means if you search with ctrl + f you it won’t show up in the results, so cannot search across the transcript.

However the performance boost should outweigh this, and a separate[ transcript wide search(and replace?) could be implemented at later stage if needed](https://reactrocket.com/post/draft-js-search-and-replace/)

Also not 100% sure how much of a performance boost `react-visibility-sensor` gives?

## unnecessary re-renders

https://github.com/bbc/react-transcript-editor/issues/159

[Rough draft of React performances optimisation research notes from various docs and blogs can be found here](https://github.com/bbc/react-transcript-editor/blob/fix-159-performance-problem/docs/notes/2019-05-16-prevent-unnecessary-re-renders-in-react.md)

Used
- [x] [`why-did-you-update`](https://github.com/maicki/why-did-you-update) lib 
- [x] [React profiler tool](https://reactjs.org/blog/2018/09/10/introducing-the-react-profiler.html) - [x] adding `console.log` in render method of components
- [x] refactored logic of some of the components communications to improve performance
- [x] use of `shouldComponentUpdate`
- [x] use of `PureComponent`

---

## Performance for over 1 hour

Tested with 
https://github.com/bbc/react-transcript-editor/issues/150

Seems better but would need testing on lower performance windows PC with clip and transcript over 1 hour (eg 2hours min)

Also important to note (as pointed out by @laurian) performance for longer content, is improved is paragraph blocks being edited are smaller (eg few lines) as opposed to half of the view height of more.

But not conclusive for now that this PR addresses the longer files performance issue. It does seem to reduce unnecessary re-renders tho.

## immutable js warnings
  `why-did-you-update` is commented out in `/demo/app.js` as it seems to slightly slows down performance when in development 

```js
if (process.env.NODE_ENV !== 'production') {
   const { whyDidYouUpdate } = require('why-did-you-update');
   whyDidYouUpdate(React, { exclude: [ /^HotKeysWrapper/ ] } );
} 
```

However When using `why-did-you-update` to test unnecessary re-renders there’s a warning



>immutable.js:4655 iterable.length has been deprecated, use iterable.size or iterable.count(). This warning will become a silent error in a future version. Error

As explained in this issue
https://github.com/facebook/draft-js/issues/950

DraftJS immutable JS dependency needs to be upgraded in DraftJs core library.

---

## Still to do

- [ ] Move `CustomEditor` into its own file form `TimedTextEditor`
- [ ] Review CSS modules files are divided with concerns properly separated.
- [ ] Capture some kind of ADR or doc note with main takeaway from these performance tweaks

----

Also addressed, small fix for 
https://github.com/bbc/react-transcript-editor/issues/147

pietrop:
Something else to think about for a separate PR is the idea of pagination. 

Context: DraftJS use case at FB is for small comments areas and Facebook notes. So not sure it has been tested agains large quantities of text or Many paragraphs (link to issue) or that they will ever have a use case to optimise for that (?).

Another option would be to reduce the amount of content in Draft at any given time, only to what is visible. 

Eg at a high level it could use a pagination approach, [similar to what described in this blog post](https://medium.com/@srulirapps/paginating-a-draftjs-editor-79b6e0f86596) (without Redux!) so that every time scrolling up or down, it moves/changes the paragraphs that are visible in the editor.

Eg the wrapper component keeps an array of paragraphs, and a range of indexes to the once visible in draftJs, then some logic handles swapping the current paragraphs... etc..

Done in this way pagination would be hidden from the users UI/UX, eg it wouldn't look like paginated text, but it could provide a considerable performance boost.

sshniro:
Hi @pietrop is the paginating issue is still pending ? Shall I give it a try with React hooks and context ?

pietrop:
Hi @sshniro,
That would be awesome, PR welcome, in the last comment I started flashing out a direction for a possible solution. You are welcome to  try that or some other approach.

I am not sure hooks and context would necessarily help with the specifics tho, as the solution might entail working around the draftJs component specific performance issues, what did you have in mind?

sshniro:
Aw I thought of handling the entire state out of draft , now only read the linked article. Ill play around with draft js and get back, thanks ! :+1:

pietrop:
Awesome! Looking forward to it! 
Any questions feel free to reach out, happy to help

pietrop:
# Before 
react-transcript-editor-unecessary re-renders
# After
react-transcript-editor-unecessary re-renders - FIX

packages/components/media-player/src/PlayerControls/index.js
```diff
@@ -18,8 +14,36 @@ import {
...
 } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+import PlaybackRate from '../PlaybackRate';
+import TimeBox from './TimeBox.js';
 
 class PlayerControls extends React.Component {
+
+  shouldComponentUpdate(nextProps) {
+    if (nextProps.currentTime !== this.props.currentTime) {
```
jamesdools: there's gotta be a better way to do these big lists of updates right 😢

packages/components/transcript-editor/index.js
```diff
@@ -56,6 +46,64 @@ class TranscriptEditor extends React.Component {
     return null;
   }
 
+  // performance optimization
```
jamesdools: In this block, we have every single part of the state accounted for and checked, plus also one Prop (mediaUrl).

I checked there weren't any other special cases, but no it's looking at every single state property. So!

Instead of doing all that, we could write:

```
  shouldComponentUpdate = (nextProps, nextState) => {
     if (nextProps.mediaUrl !== this.props.mediaUrl) return true;

     return nextState !== this.state;
  }
```

jamesdools: I don't see any difference in performance or side-effects from that. If this is alright, I think we should do the same thing in the other places (as it's adding a lot of verbose and redundant code).

packages/components/transcript-editor/index.module.css
```diff
@@ -56,15 +56,15 @@
     position: relative;
   }
 
-  .header {
+  /* .header {
```
jamesdools: if we don't need it, chuck it, or if you want to keep the class name in there for later, easier to deal with `//` comments imo.

pietrop:
Thanks @jamesdools added most of the changes in https://github.com/bbc/react-transcript-editor/pull/175.

https://github.com/bbc/react-transcript-editor/pull/175/commits/3588b264855eb36cf68ae9dda2a7968a0b75220a and https://github.com/bbc/react-transcript-editor/issues/180

Going to do a separate PR for the should component update tweaks against the develop branch, to make those changes a little clearer and easier to review. 

Closing this for now.

pietrop:
addressed should component update in this standalone PR https://github.com/bbc/react-transcript-editor/pull/182

PR #163, opened 2019-05-28
Add Trint STT Adapter

pouta:
**Describe what the PR does**    
This PR adds a STL adapter for the Trint transcription service JSON API.

Missing test, already have appropriate samples in /sample

**Additional context**    
This is a follow-up from my tweet convo with @pietrop 
The only way I could try this new changes on my own react app was by publishing my changes to a npm registry. I tried to run npm run build:component and copy /dist to my project but that did not work.


pouta:
My branch has a bunch of changes from my default editor config. When ready to review those will not be there.

pietrop:
After talking to @nigelbbc seems like the TTML could be simplified as follow
```xml
<?xml version="1.0" encoding="UTF-8"?>
<tt 
xmlns="http://www.w3.org/ns/ttml" 
xmlns:ttp="http://www.w3.org/ns/ttml#parameter" 
 ttp:timeBase="media" 
ttp:frameRate="25">
<head></head>
<body><div>
<p begin="00:00:00.020" end="00:00:06.430">Hello I can see your cat as well.<br />In fact, I can't see your</p>
<p begin="00:00:06.430" end="00:00:12.230">cat because were doing those screen share o.k.</p>
</div>
</body>
</tt>
```
Some notes from Adobe Docs:

- [Learn to work with captions in Premiere Pro](https://helpx.adobe.com/uk/premiere-pro/using/working-with-captions.html#Supportedformats)
- [Premiere Pro supported file formats](https://helpx.adobe.com/uk/premiere-pro/using/supported-file-formats.html#Supportedclosedcaptioningandsubtitlefileformats)

And It seems like it might be best to add a separate button `TTML for Premiere` along side the other `TTML` export, rather then replacing.

pietrop:
Hi @pouta 

Nicely done, I was able to try the demo app in storybook with the sample text and mp3 from the trint folder.
 
Screen Shot 2019-05-28 at 12 41 32

----

re your issue - I think it might be to do with the `package.json` missing from the `dist` folder.

Need to break this command 

```js
"publish:public": "npm run build:component && rm ./dist/package.json || true && cp package.json ./dist/package.json && rm ./dist/README.md || true && cp README.md ./dist/README.md || true && npm publish dist --access public"
```
into 
```js
"publish:prep": "npm run build:component && rm ./dist/package.json || true && cp package.json ./dist/package.json  && rm ./dist/README.md || true && cp README.md ./dist/README.md || true" 
"publish:public": "npm run publish:prep && npm publish dist --access public"
```

This way if you run `npm run publish:prep` your `dist` folder should work if you copy and paste into your parent app to try out the changes.

Let me know if that doesn't work, as if does I can incorporate it as a separate fix.

---
Unrelated to your query but connected with the upcoming RP, It be good to not have the mp3 file in the repository, and instead run a transcription [of the example ted talk](https://download.ted.com/talks/KateDarling_2018S-950k.mp4) so that there's no need for a link to an extra media file, if that makes sense?

pietrop:
Related to the `mp3` file issue we should add these to the `.gitignore`

```gitignore
#ignore all video files
*.wmv
*.mpg
*.mpeg
*.mp4
*.mov
*.flv
*.avi
*.ogv
*.ogg
*.webm

#ignore audio file
*.wav
*.mp3

#ingore subtitles files
*.srt
*.sbv
*.vtt
*.ttml
*.itt
```

pouta:
> This way if you run npm run publish:prep your dist folder should work if you copy and paste into your parent app to try out the changes.
/dist should be pasted as a component or as a node_module?

I will commit the recommended changes later.

pietrop:
with that setup `dist` most likely would go in `node_module`.

But your approach of publishing it to your npm registry is also correct, and perhaps preferable. 

You my also look into [`npm link`](https://docs.npmjs.com/cli/link) - [also explained here](https://medium.com/dailyjs/how-to-use-npm-link-7375b6219557). Not very familiar with it myself, but heard is another way to achieve similar result.

Laurian:
@pouta and how do you get that JSON from Trint? if there is a way shouldn't that be documented?

pouta:
> @pouta and how do you get that JSON from Trint? if there is a way shouldn't that be documented?

Go to https://dev.trint.com/reference and check out the Export API

pietrop:
@pouta let us know if there's still interest around this, otherwise happy to close it.

pietrop:
This PR needs resolving conflicts with current master

pietrop:
🚨
Also re-iterating the **having an mp3 file committed in the repository for this PR** might be something to watch out for, as removing large files from git, can be kind of an undertaking at times. and we don't want to add unnecessary weight to the repo with media files.

pietrop:
Closing this for now, happy to re-open it if there's interest in seeing it through

Issue #152, opened 2019-05-07
Support to Google Speech To Text transcription format


eyal13579:
any plans to support Google?

pietrop:
Hi @eyal13579 , 
At the moment we have been relying on open source contributors to create the adapters they might need, and [made a guide to help with the process](https://github.com/bbc/react-transcript-editor/blob/master/docs/guides/adapters.md). As we are only using the BBC Kaldi in our projects.

If you are interested in creating a Google STT one, and need any help to setup a PR let me know.

sshniro:
Hi @pietrop , I'm familiar with Google STT and would love  to contribute to this development. I have gone through the helper docs. I have a small clarification regarding the which approach to take. Google STT provides speaker diarization as a beta feature, so is it okay if I use the beta results for creating the sample json ?

pietrop:
Hi @sshniro, that's awesome, thanks! 

re-speaker info, sounds good to use the  beta feature for  Google STT speaker diarization.

We have been using a flexible approach in the current module, because a lot of these STT have those optional attributes, so didn't want to impose them as defaults. 

You can see in the other modules, that we check if the attribute is available, if it is we break the paragraph down based on the speaker diarization, if it isn't we break the paragraphs down based on punctuation, and add place holder names to the speaker labels. The user can then change the size of the paragraphs and speaker labels in the UI. 

for example you can see in the `/packages/stt-adapters/bbc-kaldi` we check if the attribute is present in the json ([`/packages/stt-adapters/bbc-kaldi/index.js#L57`](https://github.com/bbc/react-transcript-editor/blob/master/packages/stt-adapters/bbc-kaldi/index.js#L57)) and then we either do `groupWordsInParagraphs` (in[`/packages/stt-adapters/bbc-kaldi/index.js#L15`](https://github.com/bbc/react-transcript-editor/blob/master/packages/stt-adapters/bbc-kaldi/index.js#L15) ) or `groupWordsInParagraphsBySpeakers`( in[`/packages/stt-adapters/bbc-kaldi/group-words-by-speakers.js`](https://github.com/bbc/react-transcript-editor/blob/master/packages/stt-adapters/bbc-kaldi/group-words-by-speakers.js))

I think it be good to add support for both, creating paragraph based on speaker label as well as on punctuation, as a fallback if speaker labels are not provided. However if you want to priorities the speaker labels, and leave the punctuation one for a separate PR that's also fine.

Let me know if you have any questions.

[...]

PR #167, opened 2019-07-18
Adding speech to text adapter for Google cloud platform

sshniro:
Fix for #152 

**Describe what the PR does**    
The PR converts google speech to text response to Draft Js format. 

Completed

**Additional Context**
Google's STT response is similar to IBM's response format, so this PR follows the similar pattern for formatting the text. For example the text is broken into smaller chunks, which mostly resembles a full sentence. There for the content is not broken by punctuation as punctuation is an additional attribute which should be specifically requested from the API.


demo/select-stt-json-type.js
```diff
@@ -18,6 +18,7 @@ const SttTypeSelect = props => {
     <option value="vtt-youtube" disabled>Youtube VTT</option>
     <option value="amazontranscribe">Amazon Transcribe</option>
     <option value="digitalpaperedit">Digital Paper Edit</option>
+    <option value="gcp">Google STT</option>
```
pietrop: It be good to rename `gcp` to something more consistent with the other adapters eg `google-stt` `google-cloud-stt`, `google-cloud-platform` etc..

```suggestion
    <option value="google-stt">Google STT</option>
```

packages/stt-adapters/gcp/index.js
```diff
@@ -0,0 +1,118 @@
...
+const NANO_SECOND = 1000000000;
+
+/**
+ * attribute for the sentences object containing the text. eg sentences ={ punct:'helo', ... }
+ *  or eg sentences ={ text:'hello', ... }
+ * @param sentences
+ */
+export const getBestAlternativeSentence = sentences => {
+  if (sentences.alternatives.length === 0) {
+    return sentences[0];
```
pietrop: I guess is fair to assume the first result in the alternatives is the most accurate? and not worth filter/find the one with the highest confidence score?
(in other STT services, generally that's the case and if it's not accurate the other alternatives are generally also not good, but not sure if this applies for google STT as well?)

packages/stt-adapters/gcp/index.js
```diff
@@ -0,0 +1,118 @@
...
+  return text.trim();
+};
+
+/**
+ * GCP does not provide a nanosecond attribute if the word starts at 0 nanosecond
+ * @param startSecond
+ * @param nanoSecond
+ * @returns {number}
+ */
+const computeTimeInSeconds = (startSecond, nanoSecond) => {
```
pietrop: more of a question, I haven't seen the spec of the google STT schema, in a word like 

```json
 "startTime": {
                "seconds": "24",
                "nanos": 600000000
              },
```
I am assuming `seconds` and `nanos` need to be combined to get, in this case, the `startTime`?

packages/stt-adapters/gcp/index.js
pietrop: ah, sorry my bad, I just noticed you are actually doing a `reduce` to get the sentence with highest confidence score 👍

packages/stt-adapters/index.js
```diff
@@ -5,6 +5,7 @@ import amazonTranscribeToDraft from './amazon-transcribe/index';
 import ibmToDraft from './ibm/index';
 import digitalPaperEditToDraft from './digital-paper-edit/index';
 import createEntityMap from './create-entity-map/index';
+import gcpSttToDraft from './gcp/index';
```
pietrop:
```suggestion
import gcpSttToDraft from './google-stt/index';
```

packages/stt-adapters/index.js
```diff
@@ -41,6 +42,12 @@ const sttJsonAdapter = (transcriptData, sttJsonType) => {
     blocks = digitalPaperEditToDraft(transcriptData);
 
     return { blocks, entityMap: createEntityMap(blocks) };
+
+  case 'gcp':
```
pietrop:
```suggestion
  case 'google-stt':
```

packages/stt-adapters/gcp/index.js
sshniro: Yea @pietrop , GCP is providing seconds in the following break down , and it the second starts at exactly 24 then it return nothing for nanos attribute. The method is to handle this workflow and to compute the exact time.

sshniro:
@pietrop I have added the changes requested in the following comment.
https://github.com/bbc/react-transcript-editor/pull/167#pullrequestreview-264222584

pietrop:
Awesome, thanks @sshniro !

Out of curiosity, what's your use case for this component?

sshniro:
I was inspired by this following paper, 
https://gfx.cs.princeton.edu/pubs/Jin_2017_VTI/Jin2017-VoCo-paper.pdf

I have the problem making soo much filler words (Eg , so, and, ehh) during screen cast/ video tutorials. So I wanted to build a opensource editor for voice. Did a basic search but couldn't find an opensource equivalent. So decided to create one. And by doing a initial research I found out Google is pretty good at transcribing audio than the opensource counterparts.

So the idea is to automatically transcribe the video content and let the user to crop/replace words in the editor. The removed text should be automatically removed from the audio content as well.

Replace / delete a word is a bit easy. The paper talks about speech synthesis by using phoneme and to completely modify the words. If time permits Im planning to attempt it and see.

pietrop:
Very interesting, in a similar domain we are also working on a tool to edit audio/video interviews, at the moment is more around generating rough cuts, rather then removing filler words, but it sounds like there might be some overlap.

https://github.com/bbc/digital-paper-edit-client

you can see the demo here https://bbc.github.io/digital-paper-edit-client

The idea is that 
1. You could create an automatically generated transcript 
2. Correct it if needed using `@bbc/react-transcript-editor` - [transcript correction example in demo](https://bbc.github.io/digital-paper-edit-client/#/projects/10046281c4ad4938b7d0ae6fa9899bec/transcripts/1000cjw29xii80000ird74yb19swa/correct)
3. you can then create a programme script, and highlight/annotate your material, and or use text selection to assemble a new programme/story/paper edit - [programme script/paper-editing example in demo](https://bbc.github.io/digital-paper-edit-client/#/projects/10046281c4ad4938b7d0ae6fa9899bec/paperedits/86e938c468de4b0cb873c6e97a3a25f8)
4. and quickly review an audio/video version without needing to export
5. when done, you can export in an editing software to continue with your edit. 

if that makes sense?

sshniro:
Oh this is so cool ! Thanks for pointing me to this repository. I Will go through the issues and see if I can contribute to some features. As per the issues its more towards AWS I presume?

pietrop:
It follows a [modular architecture](https://github.com/bbc/digital-paper-edit-client/blob/master/docs/guides/modular-architecture.md) so there's a React client that in theory is not super opinionated about the backend. And the backend can be an API server or wrapped inside an electron app to package for mac, linux, and windows as desktop app.  The [`README` Project Architecture](https://github.com/bbc/digital-paper-edit-client#project-architecture) section does a better job at describing this. So yeah, larger project, and a variety of different kind of issues/tickets.

Issue #176, opened 2019-08-06
Remove UI elements via configuration / Customize UI


murezzda:
**Is your feature request related to a problem? Please describe.**
We do not use the local save function for our use case, thus it only confuses the users. It would be great to have an option to remove UI-elements if they are not needed.

**Describe the solution you'd like**
I've provided a simple PR which passes a parameter from the intialization of the transcript-editor to the player controls. This would solve my use case.

**Describe alternatives you've considered**
If other people are also interested in selectively hiding UI-elements, it might be better to have a more structured approach to customizing the UI. Maybe a configuration file which specifies certain aspects of the application.

**Additional context**
Will provide a simple PR with a straight pass-through parameter, if interest is there it would be possible to extend this to other UI-Elements with a more structured approach.

PR #175, opened 2019-07-31
Develop

pietrop:
**Is your Pull Request request related to another [issue](https://github.com/bbc/react-transcript-editor/issues) in this repository ?**      

https://github.com/bbc/react-transcript-editor/pull/166
https://github.com/bbc/react-transcript-editor/pull/160
https://github.com/bbc/react-transcript-editor/pull/144

**Describe what the PR does**    

Brings together these 3 PRs into an alpha release `@bbc/react-transcript-editor@1.0.4-alpha.2`

**State whether the PR is ready for review or whether it needs extra work**    

Work in progress, pending review by @jamesdools of those 3 PRs.

**Additional context**    

to try alpha release locally do 
```
npm install @bbc/react-transcript-editor@1.0.4-alpha.2
```

and to inspect this branch, do `npm install` first

PR #177, opened 2019-08-06
Added parameter for hiding save button, straight property pass-through

murezzda:
This PR references #176. 

**Describe what the PR does**    
The `TranscriptEditor` tag has a new parameter `showSaveButton`. This is a boolean which is passed down to `PlayerControls`, where it hides or shows the save button.

This is a example PR for this Problem which is very simple. If there is interest of developing such a feature to a greater extend, it would be better to use a more structured approach like a configuration file instead of passing down multiple parameters for each individual component.

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -262,6 +262,7 @@ class App extends React.Component {
           handleAnalyticsEvents={ this.handleAnalyticsEvents }
           title={ this.state.title }
           ref={ this.transcriptEditorRef }
+          showSaveButton= { true }
         />
 
         <label>Components Analytics</label>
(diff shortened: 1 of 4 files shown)


pietrop:
still need to discuss this with @jamesdools 
But a few thoughts for now, 

- Could this be modified to have a default as true?
- Removing the save button doesn't stop the auto save in local storage from doing it's thing. Is this going to be a problem?

jamesdools:
Agreed, @pietrop - I think it's a question on whether we do this (toggling the rendering on/off) or being able to map the save function to a customisable thing.

Also yes, if it is not rendering but still autosaving, will that cause problems when it tries to read from local storage as you boot it up.

Sidenote: as per the PR on optimising re-renders (https://github.com/bbc/react-transcript-editor/pull/175), trying to move creating new elements out of the render block.

Will pick this up soon - we have a similar problem on multiple save buttons when integrating this with larger applications.

pietrop:
@murezzda see if this PR (that has been merged) addresses your use case as well https://github.com/bbc/react-transcript-editor/issues/200

murezzda:
@pietro yes, #200 seems to address this as well. Thanks a lot!

Issue #178, opened 2019-08-06
Alert user if html5 player is unable to play file


murezzda:
**Is your feature request related to a problem? Please describe.**
We allow users to upload their own files, which - from time to time - results in audio or video files uploaded which are not compatible with the html5 video element. At the moment, this results in the play button not working, but no other response is given to the user. We would like to have an addtional indicator for the user that the uploaded file is not playable due to incompatibility with the html5 video element and not due to other problems.

**Describe the solution you'd like**
I've provided a PR which checks if a video is playing after hitting the play button. If the video is still paused, an `alert` with the Text '_The video player is not able to play the file. Please ensure the video format and codec is compatible with HTML5._' is shown. This would give the user some hints what is wrong. A detailed explanation could then be given in a FAQ or similar.

**Describe alternatives you've considered**
The solution provided could be improved by the following points:

- There might be other cases where the video is not playing after hitting the play button. There might be a better way to check for non-compatible formats.
- The alert is pretty barebone. A better indicator, e.g. a css pop-up, might be visually less confusing to the user and be able to explain the problem better.
- The provided message is probably to technical to the average user, a better explanation would be good.

pietrop:
👋 Thanks @murezzda ! 
Let me think about this and discuss it with @jamesdools .

We have tackled this problem by converting the media into HTML5 compatible audio or video after the user uploads the original media.

- [`bbc/digital-paper-edit-video-preview-converter/convert-to-video-preview/index.js
`](https://github.com/bbc/digital-paper-edit-video-preview-converter/blob/master/convert-to-video-preview/index.js)
- [`/bbc/digital-paper-edit-audio-converter/lib/convert-to-audio/index.js`](https://github.com/bbc/digital-paper-edit-audio-converter/blob/master/lib/convert-to-audio/index.js)

Out of curiosity, do you do any pre-processing before sending it to the STT service?

murezzda:
Hi @pietrop 

The option of converting the file would be the better choice, but due to resource constraints we did not implement this yet, as the speech2text service can handle the formats for itself (i.e. it does a conversion, but we can't access the converted file).

This would be a workaround for us for the moment in order to inform users that the reason the file is not playing is because the video element is unable to handle the file.

PR #179, opened 2019-08-06
Added altert which pops up if video player is unable to play the provided file

murezzda:
Related to #178.

**Describe what the PR does**    
After hitting the play button, the player checks if the video player is playing the video. If this is not the case, an `alert` is triggered informing the user that the provided file is not compatible with the html5 video player.

The given PR would be sufficient for us for the moment. Possible improvements are described in the issue page, but I can't say if I'm able to work on these in the near future.

### DIFF
--- a/packages/components/media-player/index.js
+++ b/packages/components/media-player/index.js
@@ -257,8 +257,13 @@ class MediaPlayer extends React.Component {
   };
 
   playMedia = () => {
-    this.setState({ isPlaying: true }, () =>
-      this.props.videoRef.current.play()
+    this.setState({ isPlaying: true }, () => {
+      this.props.videoRef.current.play();
+      if (!this.isPlaying()) {
+        alert('The video player is not able to play the file. Please ensure the video format and codec is compatible with HTML5.');
+        this.pauseMedia();
+      }
+    }
     );
 
     if (this.props.handleAnalyticsEvents) {


murezzda:
Hi @jamesdools, we have some problems with wav's containing audio with is mpeg1/2 encoded. I could provide you a sample file if you need it. 

I'm not sure what you mean by "load not play", how could I spot the difference?

Regarding the `canPlayType`, that would be probably the prefered solution but I'm not sure if this only checks the container or also the coded information.

pietrop:
This might be interesting via @bevand10 [HTML5 Audio — The State of Play
Tuesday, May 8th, 2012](http://html5doctor.com/html5-audio-the-state-of-play)  by @maboa

pietrop:
Another option mentioned by @bevand10  is to use the `accept` attribute in the `input` element when the users upload the file, see [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file)

Also [MDN - Media formats for HTML audio and video](https://developer.mozilla.org/en-US/docs/Web/HTML/Supported_media_formats) for a comprehensive list of HTML5 supported media.

pietrop:
Example of mime types check via @bevand10  [goldfire/howler.js/src/howler.core.js#L256](https://github.com/goldfire/howler.js/blob/master/src/howler.core.js#L256)

PR #183, opened 2019-08-14
Refactor media player

jamesdools:

**Describe what the PR does**    

Refactoring the codebase, starting with MediaPlayer. Initially making the methods and props more succinct, will add testing and optimisations later.

Going by commit:

* Fixed the font configuration in the storybook config. Realised we want Serif for titles and sans for body text. More readable now.
* Fixed the loading order of the top level components and cleaned up that code. Made ES6 changes and removed many `if videoRef !== null` checks.
* Fixed the webpack configuration - colours / globals were becoming really weird and hard to refactor. Now using scss variables at base, updated webpack with sass loaders (no extra packages needed).
* The sidebar in storybook is empty - hiding for now. Better for demo.
* Quick tidy of the demo styling and code.
* General refactor of MediaPlayer styles
  * Many were copy and pasted entirely into different breakpoints, doubling or tripling the amount of css required, and also making it very unreadable.
  * Big cut down on duplicated and redundant / or broken css.
  * Refactored into nested scss, so more readable. Was getting difficult!
  * Added some TLC, better heading spacing, line-height, and hover states for buttons.
* Removed PauseWhileTyping, Rollback and ScrollIntoView components. All unused.
* Put non components into config folder 

👍
**Additional context**


pietrop:
@jamesdools given that in the last refactor we introduced a bug https://github.com/bbc/react-transcript-editor/issues/185  (now fixed https://github.com/bbc/react-transcript-editor/pull/186) I suggest that before going ahead with this the QA doc should be updated, and QA should be run as an acceptance criteria before merging PRs that do substantial refactor.

jamesdools:
Done the QA on this - looks okay to me.

.storybook/webpack.config.js
```diff
@@ -6,36 +6,28 @@ module.exports = {
   module: {
     rules: [
       {
-        test: /\.module.css$/,
```
pietrop: does this mean we are only supporting CSS modules?

demo/index.module.scss
```diff
@@ -0,0 +1,142 @@
+body {
+  margin: 0;
+  padding: 0.5em;
+}
+
+.container {
+  height: 100vh;
+  font-family: ReithSerif, Fallback, sans-serif;
```
barneyboo: I don't think `Fallback` will resolve to anything - is this meant to be a reference to a mixin?

demo/index.module.scss
```diff
@@ -0,0 +1,142 @@
...
+.demoButton {
+  background-color: #74b567 !important;
+
+  &:hover {
+    background-color: #528946 !important;
+  }
+}
+
+/* desktop */
+@media (min-width: 767px) {
```
barneyboo:
```suggestion
@media (min-width: 768px) {
```
Breakpoints can go a bit weird if the min/max are the same - better for the `min-width` to be `max+1`

packages/components/media-player/src/PlayerControls/TimeBox.js
```diff
@@ -1,20 +1,17 @@
...
 
-  shouldComponentUpdate(nextProps) {
-    if (nextProps !== this.props) {
-      return true;
-    }
+class TimeBox extends React.Component {
 
-    return false;
+  shouldComponentUpdate = (nextProps) => {
+    return nextProps !== this.props;
```
barneyboo: I think checking for object equality here will always return `true` (since it only checks that they're the same reference in memory). Better to compare on the relevant props that might have changed between renders.

packages/components/media-player/src/PlayerControls/index.js
```diff
@@ -19,18 +17,14 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
...
 
+import style from './index.module.scss';
+
 class PlayerControls extends React.Component {
 
   shouldComponentUpdate = (nextProps) => {
-    if (nextProps !== this.props) return true;
-
-    return false;
+    return nextProps !== this.props;
```
barneyboo: Same comment as above ^ re: checking props equality

demo/index.module.scss
jamesdools: Ah yeah, I'm never quite sure on this. So we should move to 768?

demo/index.module.scss
jamesdools: Good catch - I followed this from somewhere (I'm sure int-gel-matter or a component library) but I can't find it anymore - it's only us doing it!

So yeah - I think this can be removed 😅

packages/components/media-player/src/PlayerControls/TimeBox.js
jamesdools: Hmm. Yeah this was to improve the rendering performance but without cluttering the code too much. Is there a working one-liner for this?

jamesdools: Thinking either the lodash `isEqual` or perhaps this:
https://www.npmjs.com/package/react-fast-compare

have you used it before?

PR #189, opened 2019-08-16
Bugfix: Editor grabs focus on realignment even if editor not in focus

murezzda:
**Describe what the PR does**    
After word realignment, the editor needs to set the focus again on the previous focus position. At the moment, this is also the case if the user has the focus not in the editor. In that case, the editor grabs the current focus away from whatever the user is doing, which is quite annoying.

This PR fixes this problem by checking if the editor has focus at the time. If not, the focus will not be set back to the previous focus position

**Additional context**    
The bug can be observed in the demo application: start the player and open the drop-down for importing or exporting transcripts. The focus will switch back to the player, making it hard to select a new setting.

### DIFF
--- a/packages/components/timed-text-editor/index.js
+++ b/packages/components/timed-text-editor/index.js
@@ -129,26 +129,32 @@ class TimedTextEditor extends React.Component {
     // Re-convert updated content to raw to gain access to block keys
     const updatedContentBlocks = convertToRaw(updatedContent);
 
-    // Build block map, which maps the block keys of the previous content to the block keys of the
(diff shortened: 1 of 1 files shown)


pietrop:
Looks good, thanks @murezzda, was also able to verify locally both on demo and branch the difference.

pietrop:
I've also updated the changes to `@bbc/react-transcript-editor@1.0.7` and the demo in storybook accordingly.

murezzda:
Hi Pietro. Thanks for the promt response!

PR #193, opened 2019-09-17
Adding test to timed text editor

emettely:

Not from this issue, but DPE has a dependency on react-transcript-editor and we want to start adding tests to it.
https://github.com/bbc/digital-paper-edit-electron/issues/4

**Describe what the PR does**    
Adding the most basic test to timed-text-editor and also removing CustomEditor because it's a wrapper component for one single function - we don't really need it.

Ready

**Additional context**


pietrop:
> removing CustomEditor

You might want to check this change against this issue https://github.com/bbc/react-transcript-editor/issues/159 (which is the reason why it was initially separated into its own component)

see https://github.com/bbc/react-transcript-editor/pull/160 for more details.

TL;DR: using react inspector tool `highlight updates` you can see if removing `CustomEditor` as separate component increases the number of unnecessary re-renders. You'll have to judge it based on colour.

emettely:
A very good catch, the re-renders definitely don't happen for the original ones.

emettely:
https://github.com/facebook/react/issues/16437#issuecomment-524892514

emettely:
Added back the Editor as a memoized react component, and now the re-renders are gone. 
Screen Shot 2019-09-17 at 16 48 32

PR #201, opened 2019-10-03
Fix #200 component save

### LINKED ISSUE #200: Component save mechanism
**Describe the bug**

At the moment the component uses the url param as a key to save in local storage.
However when you come back to a transcript if it is present in local storage with the current logic that version takes priority over the server side one.

This could be problematic if there are multiple people (asynchronously) editing the transcript and the server version has moved passed the local version. As the local version will be left behind.

**To Reproduce**
see above description

**Expected behavior**

To have a better for the component to handle saving that doesn't introduce bugs.

**Screenshots**

_NA_

**Additional context**

https://github.com/bbc/digital-paper-edit-electron/issues/4
https://github.com/bbc/react-transcript-editor/pull/177

pietrop:

https://github.com/bbc/react-transcript-editor/issues/200

**Describe what the PR does**    

Option 1 in https://github.com/bbc/react-transcript-editor/issues/200#issuecomment-537869007

>In current master to save to server the logic is already outside of the component using ref.
>
>The suggestion is to revisit that, and move the logic for local storage saving outside of the component as an optional module available at user discretion.
>
>This also leaves to the user(developer) the responsibility for the logic on what to do when local storage maxes out. As well as whether they'd rather use some other form of local storage, eg indexDb.

Adds the following attributes to `TranscriptEditor`

| Attributes            | Description                                                                                                             | required |   type    |
| :-------------------- | :---------------------------------------------------------------------------------------------------------------------- | :------: | :-------: |
|`handleAutoSaveChanges`| returns content of transcription after a change                                                                         |    no    |  Function |
| autoSaveFormat        | specify the file format for data returned by `handleAutoSaveChanges`,falls back on `sttJsonType`. or `draftjs`          |    no    |  string   |

- `fileName` is no longer used as key to store and retrieve from local storage, but is still in use when exporting transcript in various formats. and it's still optional.
- logic in `onChange` to auto save in `TimedTextEditor` was trigger a bit too often, as it was not inside the check to see if the content of draftJs had changed (`this.state.editorState.getCurrentContent() !== editorState.getCurrentContent()`) so moved it to only be triggered if content has been modified (inserted, deleted etc..)
- updated `TranscriptEditor` storybook ([`http://localhost:6006/?path=/story/transcripteditor--default`](http://localhost:6006/?path=/story/transcripteditor--default))
- `setState` is async so changed the logic to use result of `updateTimestampsForEditorState` when returning data for autoSave.
- added an auto save text area for inspection and troubleshooting in demo app in storybook next to the analytics text area
- moved function to save and retrieve from local storage in `local-storage.js` under the demo app, consider whether it's work, in a separate PR, showcase how local storage save would work in the demo app etc.. I'd say low priority tho?

Ready for review

**Additional context**    

related Issue
https://github.com/bbc/react-transcript-editor/issues/176
PRs
https://github.com/bbc/react-transcript-editor/pull/177
https://github.com/bbc/react-transcript-editor/pull/199

Once this PR is merged it be good to 
 - [ ] update the the storybook on gh pages 
 - [ ] update npm


demo/app.js
```diff
@@ -156,126 +179,176 @@ class App extends React.Component {
     this.setState({ fileName: value });
   };
 
+  handleAutoSaveChanges = newAutoSaveData => {
+    console.log("handleAutoSaveChanges", newAutoSaveData);
+    const { data, ext } = newAutoSaveData;
+    this.setState({ autoSaveData: data, autoSaveExtension: ext });
```
emettely: If the autoSaveExtension is stored here (and assuming that's the correct one), would you actually still want autoSaveFormat?

pietrop: Maybe the naming needs to be changed to be more clear `autoSaveExtension` refers to the file extension. eg `.srt` or `.json` file while `autoSaveFormat` is `draftjs` or `digitalpaperedit` etc..  if that makes sense?

demo/app.js
```diff
@@ -21,28 +29,46 @@ class App extends React.Component {
...
       spellCheck: false,
-      sttType: 'bbckaldi',
+      sttType: "bbckaldi",
       analyticsEvents: [],
-      title: '',
-      fileName: ''
+      title: "",
+      fileName: "",
+      autoSaveData: {},
+      autoSaveFormat: "draftjs",
```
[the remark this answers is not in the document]
pietrop: in the actual `TranscriptEditor` component is treated like props. Here in demo I added it in state so that I could display it here [demo/app.js#339](https://github.com/bbc/react-transcript-editor/pull/201/files#diff-b705f9ad2d951e512083753f54b4cad2R339) as well as if we want to add an option in the demo navbar to change that setting and see different on save formats etc..

demo/app.js
emettely: ah right, yeah that makes sense. ok, if you can - can you quickly rename `autoSaveFormat` to something like `autoSaveObjectType`?

Issue #141, opened 2019-04-09
Updating a speaker name should update all speakers with that name


philmcmahon:
**Is your feature request related to a problem? Please describe.**
Currently, updating a speaker name just updates that block.

**Describe the solution you'd like**
I'd like it to update every occurence of that speaker name in the document - so every instance of 'Speaker 0' can become 'Hugh Dennis' 

**Additional context**
We're hoping to plug the output of this tool into the frontend of our website at some point, it would be good to avoid people having to do find+place tidy up later. I've had a go at implementing this [here](https://github.com/bbc/react-transcript-editor/compare/master...philmcmahon:update-all-speakers) without success yet (the UI doesn't update) but thought I'd check in before spending more time on it to find out if this is a feature you'd welcome?

pietrop:
Thanks @philmcmahon, yes this would be a good feature to add.
@scentless-apprentice was also interested in this. 

There's two things/features to consider in this context

1. Is the one you mentioned, of being able to batch change all speakers. You would have to work out
   - if there are any edge cases to be mindful of where you wouldn't want this to happen
   - If it's an option given to the users
   -  if it's something that only happens when the user changes the "default" speaker label
2. The other was suggested by @scentless-apprentice, which is when you click on a speaker label, it could also have an autocomplete dropbdown of the speakers that have already been used. (This is also useful for when the speaker identification, is either not present or has failed)

These might either be two separate features, or integrated to be part of the same solution, depending on the approach taken, but thought I'd connect you in the interest of non duplicating work 😄 

Had a quick scan through your branch, there's a chance you might need to force a re-render (`forceRenderDecorator`) let me know if that doesn't work tho.

bevand10:
My view is that you offer the following initially:

* Capture the name change.
* Then ask the user “You’re about to replace ‘Speaker0’ with ‘Dave Bevan’. Set the dialog buttons to ‘Cancel’, ‘Yes, replace all 27 occurrences’ and ‘Yes, but just this instance.’

If n==1 then you’d only need to prompt ‘Cancel’ & ‘Yes’.

That way you are being clear about the impact of a bulk change.

Thoughts?

pietrop:
👋 @philmcmahon just following up to see if there was any progress on this? or whether you might have any questions?

philmcmahon:
[...]

PR #208, opened 2019-10-07
Fix 141 updating speaker names

pietrop:

https://github.com/bbc/react-transcript-editor/issues/141

**Describe what the PR does**    

Improves the experience of correcting speaker labels 

Work in progress

**Additional context**    

_NA_


pietrop:
can change
- [x] one speaker at a time, 
- [x] all speakers with same label name.

 
speaker-change-1


- [x] and it triggers auto save callback on speaker/s change

packages/components/timed-text-editor/CustomEditor.js
```diff
@@ -0,0 +1,66 @@
...
+        handleAnalyticsEvents: this.props.handleAnalyticsEvents,
+        isEditable: this.props.isEditable
+      }
+    };
+  };
+
+  shouldComponentUpdate(nextProps) {
+    // https://stackoverflow.com/questions/39182657/best-performance-method-to-check-if-contentstate-changed-in-draftjs-or-just-edi
+    if (nextProps.editorState !== this.props.editorState) {
+      console.log('it should update??!')
```
emettely:
```suggestion
```

packages/components/timed-text-editor/WrapperBlock.js
```diff
@@ -55,14 +75,38 @@ class WrapperBlock extends React.Component {
...
+        return true;
+      }
+  }
 
+  handleOnClickEdit = () => {
+    const oldSpeakerName = this.state.speaker;
+    const newSpeakerName = prompt('New Speaker Name?', this.state.speaker);
     if (newSpeakerName !== '' && newSpeakerName !== null) {
       this.setState({ speaker: newSpeakerName });
+      const isUpdateAllSpeakerInstances = confirm(`Would you like to change all occurrences ${oldSpeakerName} of with ${newSpeakerName} or just this one?\n\n pick OK for all, or CANCEL for just this one.`);
```
emettely: So playing around with this, I wonder if it's possible to add 3 options rather than OK (all) and CANCEL (one) because it can be misleading, even with the text. Some flashbacks on error prone dialogue windows :D
I would expect something like:

"Would you like to change"
all, one, cancel (default - e.g. when you press enter)

This might mean changing from a single `confirm` to a nested confirm` see (https://stackoverflow.com/a/9091451), because you cannot add more than yes/no to a `confirm`.

pietrop: yeah, fair point, thanks, will have a look into it

pietrop: Ok, I played around with it, I am not sure that the nested prompt would help with clarity. 

I think it might be more a case of better UX writing of the instructions?
eg 
- Speaker name? ok/cancel
- Would you like to apply to this change to all instances? ok/cancel

What do you think?

emettely: the nested confirm should show two prompts of what you suggested above in succession (speaker name confirm and change all confirm) - and i think the phrasing of what you suggested is good!

emettely: My bad - you are correct in saying that it is the same set up.
It's mad that the JS cannot be changed (saw [this](https://stackoverflow.com/questions/22885897/javascript-rename-confirm-buttons)... Since I strongly believe in using the safer action as the primary action:
https://uxmovement.com/buttons/visual-weight-of-primary-and-secondary-action-buttons/

I guess it's either you implement your own (which might be better in the longer run, but time consuming), or you just use what they give you. At this stage, until proven that the default button has created problems for the user, I am ok with going with the latter.

emettely:
```suggestion
      const isUpdateAllSpeakerInstances = confirm(`Would you like to replace all occurrences of ${oldSpeakerName} with ${newSpeakerName}?`);
```

emettely: One feature that would be nice to have with such actions is an UNDO feature.

PR #209, opened 2019-10-08
Speaker dropdown selection

pietrop:
**Is your Pull Request request related to another [issue](https://github.com/bbc/react-transcript-editor/issues) in this repository ?**      

Improving https://github.com/bbc/react-transcript-editor/pull/208, https://github.com/bbc/react-transcript-editor/issues/141 - https://github.com/bbc/react-transcript-editor/issues/141#issuecomment-481361315

**Describe what the PR does**    

>when you click on a speaker label, it could also have an autocomplete dropbdown of the speakers that have already been used.

**State whether the PR is ready for review or whether it needs extra work**    

Work in progress

**Additional context**    

_NA_

PR #210, opened 2019-10-08
Speaker dropdown select -v2

pietrop:
### Is your Pull Request request related to another [issue](https://github.com/bbc/react-transcript-editor/issues) in this repository ?

Improving https://github.com/bbc/react-transcript-editor/pull/208, https://github.com/bbc/react-transcript-editor/issues/141 - https://github.com/bbc/react-transcript-editor/issues/141#issuecomment-481361315

_alternative approach to https://github.com/bbc/react-transcript-editor/pull/209, with no external dependencies introduced._

### Describe what the PR does

>when you click on a speaker label, it could also have an autocomplete dropbdown of the speakers that have already been used.

### State whether the PR is ready for review or whether it needs extra work 

Ready for review + one small draftJs bug https://github.com/bbc/react-transcript-editor/pull/210#issuecomment-539963878

### Additional context    

Initially considered to do an auto complete dropdown as suggested by @scentless-apprentice 
 and tried to use [react-select](https://react-select.com) in this PR https://github.com/bbc/react-transcript-editor/pull/209  to help with that. But then decided to scrap it and use html select element directly, combining it with features from https://github.com/bbc/react-transcript-editor/pull/208

##### Other considerations 
- auto complete is neat, but perhaps out of scope for a version 1. Based on anecdotal evidence from user interview, it's unlikely the number of speakers per transcripts will exceed 5 at most for the majority of use case, with some edge cases where it could go up to 10 (we are talking about audio/video interview for factual/documentary production).
- There is also further complications in making content of `WrapperBlock` editable, I think we set the focus of cursor to be on the text of the editor only, and forgot how we did it, to be able to look into how to tweak that. However using select combined with `prompt` that is not necessary for a first version.


pietrop:
Feature checklist 
- [x] Can change a speaker by creating a new speaker
- [x] Can change all speakers (with a certain label) by creating a new speaker
- [x] Can change a speaker by choosing from dropdown selection
- [ ] Can change all speakers (with a certain label)  by choosing from dropdown selection <-- change does not re-render

pietrop:
All seems good, the features seem to be working as expected.


However when using select inside the `WrapperBlock` get an error 
```
Uncaught TypeError: Cannot read property 'nodeType' of null
    at editOnInput
```
I think this is a draftJs problem, see https://github.com/facebook/draft-js/issues/1085

Screenshot 2019-10-09 at 12 38 29


I've  also updated draftJs to the latest version as the console kept complaining that current version of draft uses an older version of React.

pietrop:
Raised issue https://github.com/facebook/draft-js/issues/2204

emettely:
Sorry capturing GIFs don't work with licecap (probably to do with Firefox...) :( so having to use old school image capturing.

In a more narrow set up (960) the drop-down seems to disappear.

# 1 Click speaker



# 2 Drop down menu disappears



# 3 ... but not really
Random guess-clicking shows it's still there.

emettely:
Just capturing this: the ordering of numbers is not the greatest, but this is the nature of treating numbers as strings. As @pietrop said, if people are thinking to only have up to 5 speakers (also assuming that they have real names not numbers), it's not a problem.

Screen Shot 2019-10-10 at 16 08 34

emettely:
> Raised issue [facebook/draft-js#2204](https://github.com/facebook/draft-js/issues/2204)

This might be the answer: 
https://draftjs.org/docs/advanced-topics-block-components.html#recommendations-and-other-notes

PR #217, opened 2020-01-01
Fix Performance issue for long videos.

aneesijaz:

ISSUE: [https://github.com/bbc/react-transcript-editor/issues/150](Performance hit for media over 1 hour #150)

**Describe what the PR does**    

Added react-visibility-sensor and modified wrapperBlock.js file to load only the visible area which can be helpful for the long videos with long transcript.

Ready

**Additional context**    

This is my 1st PR so please let me know if I did anything wrong and I will amend my mistake.
Thanks.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13700,6 +13700,14 @@
         "react-lifecycles-compat": "^3.0.4"
       }
     },
+    "react-visibility-sensor": {
+      "version": "5.1.1",
+      "resolved": "https://registry.npmjs.org/react-visibility-sensor/-/react-visibility-sensor-5.1.1.tgz",
+      "integrity": "sha512-cTUHqIK+zDYpeK19rzW6zF9YfT4486TIgizZW53wEZ+/GPBbK7cNS0EHyJVyHYacwFEvvHLEKfgJndbemWhB/w==",
+      "requires": {
+        "prop-types": "^15.7.2"
+      }
+    },
     "reactcss": {
       "version": "1.2.3",
       "resolved": "https://registry.npmjs.org/reactcss/-/reactcss-1.2.3.tgz",
@@ -15443,6 +15451,16 @@
                 "pac-resolver": "^3.0.0",
                 "raw-body": "^2.2.0",
                 "socks-proxy-agent": "^4.0.1"
+              },
+              "dependencies": {
+                "https-proxy-agent-snyk-fork": {
+                  "version": "git://github.com/snyk/node-https-proxy-agent.git#5e86ccb682d0c833c8daa25ee6f91c670161cd66",
+                  "from": "git://github.com/snyk/node-https-proxy-agent.git#fix/https-agent-vuln",
+                  "requires": {
+                    "agent-base": "^4.3.0",
+                    "debug": "^3.1.0"
+                  }
+                }
               }
             },
             "pac-resolver": {
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "react-fast-compare": "^2.0.4",
     "react-keyboard-shortcuts": "^1.1.3",
     "react-simple-tooltip": "^2.6.1",
+    "react-visibility-sensor": "^5.1.1",
     "sbd": "^1.0.15",
     "smpte-timecode": "^1.2.3"
   },
--- a/packages/components/timed-text-editor/WrapperBlock.js
+++ b/packages/components/timed-text-editor/WrapperBlock.js
@@ -7,7 +7,7 @@ import {
   convertFromRaw,
   convertToRaw
  } from 'draft-js';
-
+ import VisibilitySensor from 'react-visibility-sensor';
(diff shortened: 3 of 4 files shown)


pietrop:
Thanks soo much for looking into this @aneesijaz !

Trying out this PR, here's a few things I noticed, testing with a 2 hour video

## Local storage
Not necessary related to your PR, but in the demo app in storybook, while testing saving changes to local storage causes problems 

Screenshot 2020-01-03 at 11 46 54

So I commented out line 10, as this can be addressed as separate issue, more to do with how the parent application deals with the data etc..
```js
// localStorage.setItem(`draftJs-${ mediaUrlName }`, JSON.stringify(data));
```
## Shaking
This doesn't happen all the time, but some time, while playing the text starts to shake
shaking

## Typing while playing
If I start typing while it's playing, there's a delay before the text shows up. And then it gets unresponsive for a bit afterwards, did you come across this as well?

aneesijaz:
ok so the other problem of storage exceeds is not related to this of course.
And the shaking problem is fixed. just had to remove the min height from the `loadingPlaceholder`.
Tested on my end and not shaking anymore (Yes it was shaking a lot before).
Let me know if there is any other issue.

aneesijaz:
And the issue that the editor hangs is I guess also not related to this. Because I had same experience with the visibility sensor and without the visibility sensor they both worked the same (some time's a little bit lag...)

pietrop:
Thanks for the changes @aneesijaz 

yeah, the point of issue https://github.com/bbc/react-transcript-editor/issues/150 was to see if things like visibility sensor could remove the lag.  As well as trying to narrow down what's causing it.

aneesijaz:
the lag for what the issue was a huge one.. and that was because the draftjs rendered block have a huge dom tree which cause memory to overload when all the transcript is loaded at the same time.
But with this visibility sensor module the only visible screen is  being rendered which makes a huge difference in performance.
And for the lag I think you have not tried it after the last commit , I think it is a small lag that may be considered non-existent.
I would suggest testing it for same two hours video with and without the visibility sensor, this will show the difference more clearly.
Thank you.

pietrop:
Awesome, thanks, yes I haven't tried after the last commit, will try and let you know as soon as I get a chance

jamesdools:
Hi @aneesijaz, thanks for the PR!

I just tried it locally and ran into an issue. 
Screen Shot 2020-01-21 at 16 48 05

When _Scroll Sync_ is turned on from the options, and say you click to further in the media via the progress bar, it will give you this.

Digging into the code it's because we use the querySelector to grab the word right word - but it won't find it in the DOM.

As below, I wrapped an `if (currentWordElement)` to stop the breaking error, but still need it to somehow navigate the right element even though it's not visible.

```
    if (currentWord.start !== 'NA') {
      if (this.props.isScrollIntoViewOn) {
        const currentWordElement = document.querySelector(`span.Word[data-start="${ currentWord.start }"]`);

        if (currentWordElement) {
          currentWordElement.scrollIntoView({
            block: 'nearest',
            inline: 'center'
          });
        }
      }
    }

```

I'm sure there's a way with react-visibility-sensor but haven't been able to dig through the library yet. Any thoughts?

jamesdools:
On a sidenote: in the WrapperBlock I changed "Loading" to a simple "..."

And added an offset in the props of the visibility sensor:
```
      <VisibilitySensor
        onChange={ this.visibilityChanged }
        containment={ containmentDOMRect }
        offset={ { top: -1200, bottom: -1200 } }
        partialVisibility
        scrollCheck
      >
```

I found this gives you a buffer zone of about 2 screens-worth of height (~600px for the editor)!
This means when scrolling fast through the editor, you don't see so much loading of the blocks, which to me is quite jarring.

aneesijaz:
I will look into that scroll sync , may be we can find any other method of doing that scroll. (I was thinking of adding those attributes of timings in the loading div and then when that the page is scrolled then the loading should be automatically converted back to the wrapper object.)

pbirsinger:
Wow - would love to have this PR merged! Any blockers still?

I'm finding a pretty sizable perf hit even starting around 45 mins. Does the input stt type make any difference?

pietrop:
Hi @pbirsinger 
I think this PR is not fully solving the issue https://github.com/bbc/react-transcript-editor/issues/150 so until then best to hold off.

The input stt type should not make a difference as they all get converted to draftJs json. 

But yeah, help wanted in figuring out what is the exact cause of the performance issue.

pbirsinger:
Hi @pietrop - thanks for the fast response. Do you think this PR provides any perf gain now but is buggy, or just does not significantly improve performance?

pietrop:
When I last tried it, it didn't provide any significant improvement eg with a 2 hour video. But it was a while back, you are welcome to test it out locally, and do a giff/screen recording if you want to try it out

pbirsinger:
Just tried it out, and still extremely laggy on longer videos.

Issue #247, opened 2020-10-02
Bug: unexpected scroll behavior after edit on long transcript


shea12:
**Describe the bug**
While editing a long transcript, the transcript will automatically scroll to the bottom of the current block after autosaving.

**To Reproduce**
1. Begin editing a long transcript (see gif below)
2. Stop editing, wait for autosave
3. After autosave completes, the transcript will jump to the end of the edited block

**Expected behavior**
The transcript should not scroll unexpectedly.

**Screenshots**
![react-transcript-editor-scroll-bug](https://user-images.githubusercontent.com/14244244/94939919-6cb0d380-0498-11eb-907b-dba2d3546e1d.gif)

**Desktop**
 - OS: macOS
 - Browser: (all) chrome, safari, firefox
 - Version: latest

mcwhitaker:
This happens on Windows machines as well, latest version.

jamesdools:
Hey @shea12 & @mcwhitaker!

Sorry we didn’t see that issue - it's been a long while since we've touched RTE and have been on other projects (but we’re really trying to carve some time out in the future for it).

@mcwhitaker mentioned you were keen to explore a fix yourself in the meantime - more than welcome to!

The code is very messy and hacky for sure so I imagine it'd be a real struggle haha. Lots of React things have moved on since we prototyped this (we are a Lab after all!)

We're aware of a few bugs to do with focus / cursor position and they are definitely the most jarring and experience-impeding parts to editing a transcript. This is handled in Draft.js by the `SelectionState` ([Draft docs link](https://draftjs.org/docs/api-reference-selection-state/)). If you've not dealt with Draft.js before, I gave a [quick overview in a talk](https://www.youtube.com/watch?v=E3mEUmwjxgk&t=13m&ab_channel=ReactConfAU) a while back that may or may not be helpful!

Once you start typing in the editor, all things come through this entry point [here in the `TimedTextEditor`](https://github.com/bbc/react-transcript-editor/blob/master/packages/components/timed-text-editor/index.js#L61) component.

Most of these bugs tend to happen once we autosave and attempt to re-align timecodes we have (in lines 96-97). It kicks off quite a few bulky operations but that's where to start I think.

We _were_ thinking of reducing / turning off the autosave at some point as it does tend to cause problems.

Hope that helps if you were keen to dig into it but no worries as it is a headache!

On a sidenote though - would be great to know more about your transcription use case - and might be useful if we ever get round to a V2!

[...]

PR #248, opened 2020-10-23
Fix for issue #247 "unexpected scroll behavior"

shea12:
Yes, issue #247 

**Describe what the PR does**    
This PR removes a call to setState inside the time text editor component's updateTimestampsForEditorState method because it causes the scroll bug in the issue linked above. In testing, we did not find any adverse effects of removing the call to setState.

Ready!

**Additional context**
Below is a screen recording of the fix to compare with the screen recording of the bug in the issue linked above.
react-transcript-editor-fix

### DIFF
--- a/packages/components/timed-text-editor/index.js
+++ b/packages/components/timed-text-editor/index.js
@@ -154,7 +154,7 @@ class TimedTextEditor extends React.Component {
         newEditorState,
         selection
       );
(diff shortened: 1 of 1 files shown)


emettely:
Thanks @shea12 for the PR :)

emettely (approved):

PR #249, opened 2020-11-25
Timestamp Sync Bug

shea12:
No

**Describe what the PR does**    
This PR removes the call to the interpolationOptimization method in stt-align-node. The reason for removing this call is that interpolationOptimization causes an issue with the transcript timestamps aligning correctly with the audio/video. We currently use react-transcript-editor in a STT correction workflow. In the course of correcting a transcript we discovered that the transcript and audio/video slowly get out of sync with each other. The culprit seems to be interpolation optimization - as editing takes place, it erroneously shifts each word's start time. This issue is most noticeable in transcripts where STT failed to pick up some chunks of speech.

Ready

**Additional context**    
It is possible that this fix is specific to our use case and not applicable to the wider audience of the transcript editor. If this is the case, I could add to this PR a user-facing option in the settings panel to allow interpolation optimization to be toggled on/off.

### DIFF
--- a/packages/components/timed-text-editor/UpdateTimestamps/stt-align-node.js
+++ b/packages/components/timed-text-editor/UpdateTimestamps/stt-align-node.js
@@ -100,14 +100,13 @@ function adjustTimecodesBoundaries(words) {
 }
 
 function interpolate(wordsList) {
-  const words = interpolationOptimization(wordsList);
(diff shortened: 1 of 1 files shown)


emettely (approved):
Thanks for spotting this bug!

PR #265, opened 2021-06-24
analytics for how does this work hover

sarahrainbow:
part of the work to add analytics for DPE, it fixes https://github.com/bbc/digital-paper-edit-firebase/issues/219 

**Describe what the PR does**    
* when in the transcript editor, hovering on 'how does this work' header element will trigger a handleAnalyticsEvents function

RFR

NOTE:  you won't notice analytics firing in the networks tab for this work


allishultes (approved):
lgtm!

